PR #10, opened 2021-09-09
feat: drop instead of wrap empty sections

rofe:
I think we should drop empty sections (e.g. left over from a metadata block) to avoid unexpected spacing from margins or paddings.

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -69,7 +69,7 @@ export function toClassName(name) {
  */
 function wrapSections($sections) {
   $sections.forEach(($div) => {
-    if (!$div.children.length) {
+    if ($div.childNodes.length === 0) {
       // remove empty sections
       $div.remove();
     } else if (!$div.id) {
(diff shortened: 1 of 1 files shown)


scripts/scripts.js
```diff
@@ -69,7 +69,10 @@ export function toClassName(name) {
  */
 function wrapSections($sections) {
   $sections.forEach(($div) => {
-    if (!$div.id) {
+    if (!$div.children.length) {
```
kptdobe, line 72: It is probably valid because all text should be wrapped inside `<p>` tags but according to the doc:`Element.children includes only element nodes. To get all child nodes, including non-element nodes like text and comment nodes, use Node.childNodes.`. https://developer.mozilla.org/en-US/docs/Web/API/Element/children
You might remove a div with a text node.

kptdobe (approved):

scripts/scripts.js
rofe: You're right: I should use `childNodes` instead.

davidnuescheler:
if we intend to do that on the server, i think that this would be great.

davidnuescheler (approved):

PR #11, opened 2021-09-10
feat: .hlxignore config

rofe:


.hlxignore
```diff
@@ -0,0 +1,7 @@
+.*
+*.md
+*.json
+*.yaml
```
tripodsan, line 4:
```suggestion
```
we need those :-)

tripodsan (approved):

PR #34, opened 2021-12-16
chore: updating block loader

davidnuescheler:
only vaguely related, i think we should have a default fstab in here that points to example content (read only) and i think that we could allow copying which would make it easy for someone to get started.


rofe (changes requested):
LGTM, apart from the linting errors

davidnuescheler:
@rofe & @kptdobe i updated a few more things and changed the sequence, and i am not sure why the test fails. it may very well be that i am doing something wrong.

scripts/scripts.js
```diff
@@ -122,6 +169,7 @@ function decorateBlocks($main) {
  * @param {string} blockName name of the block
  * @param {any} content two dimensional array or string or object of content
  */
+// eslint-disable-next-line no-unused-vars
```
kptdobe, line 172: Really needed ?

fstab.yaml
```diff
@@ -1,2 +1,2 @@
-# mountpoints:
-#   /: <shared-folder-url>
\ No newline at end of file
+mountpoints:
+  /: https://drive.google.com/drive/u/0/folders/1MGzOt7ubUh3gu7zhZIPb7R7dyRzG371j
```
kptdobe, line 2: You probably want to revert this one...

scripts/scripts.js
```diff
@@ -152,29 +200,43 @@ function buildBlock(blockName, content) {
...
 
 /**
  * Loads JS and CSS for all blocks in a container element.
  * @param {Element} $main The container element
  */
-async function loadBlocks($main) {
-  $main
-    .querySelectorAll('div.section-wrapper > div > .block')
-    .forEach(async ($block) => loadBlock($block));
+export function loadBlocks($main) {
```
kptdobe, line 236: Missing the JS doc for  `return`.

kptdobe (changes requested):

fstab.yaml
rofe: @kptdobe I think this is intentional and part of the new first run experience @davidnuescheler had in mind: immediately show some content and guide the developer on what to do next.

kptdobe: ah yes, this even the PR description :)

scripts/scripts.js
rofe: This function is no longer used by default, but could theoretically be used to create synthetic/auto blocks. But for that purpose, it would need to be exported...

rofe (approved):
Add an `eslint-ignore-next-line` statement for the 2 `console.log` statements (or comment them out?)

rofe:
>  i am not sure why the test fails. it may very well be that i am doing something wrong.
Never mind that, i think we should switch to web test runner anyway.

scripts/scripts.js
davidnuescheler: i think this is a function that i think we should encourage to use, possibly supported with an example. i think one of the most typical examples would be a hero/marquee section that is turned into a hero block. i have come to believe that somehow the templates or page types have a close connection with autoblocking.
meaning that if you have eg. a `category` page or a `article` page or similar those would create different auto blocks from the default content. somehow feels like an intuitive concept but i am not sure how to marry the two...
i think for now we could possibly just use an `h1` with a preceding `picture` and turn that into a `marquee` block as an example.

scripts/scripts.js
davidnuescheler: done

kptdobe (approved):

PR #40, opened 2022-01-05
feat: sequential block loader and section reveal

davidnuescheler:
https://sequential-block-loader--helix-pages-starter--adobe.hlx3.page/

vs. 

https://main--helix-pages-starter--adobe.hlx3.page/


davidnuescheler:
this adds a more predictable sequential  approach to block loading. i think it may trade performance against vs. simplicity for the dev, as a dev can rely on a specific order of execution.
i wasn't able to see any impact to LH score in the trivial case.

Screen Shot 2022-01-05 at 12 12 40 PM
Screen Shot 2022-01-05 at 12 13 33 PM

also this PR adds a sequential reveal of `sections` which should remove all `LCP` complexity from the developer, and still provide a fast appearance of FCP and LCP where possible on a per section granularity.
this may suggest a pattern to isolate LCP relevant items into smaller sections, but i am not sure that this is problematic and it is definitely a lot simpler than wrestingly with loading sequences and trying to isolate above the fold css into `styles.css`

scripts/scripts.js
```diff
@@ -157,6 +157,38 @@ export function decorateBlock(block) {
...
+
+/**
+ * Updates all section status in a container element.
+ * @param {Element} $main The container element
+ */
+function updateSectionsStatus($main) {
+  const sections = [...$main.querySelectorAll(':scope > div.section-wrapper')];
+  for (let i = 0; i < sections.length; i += 1) {
+    const section = sections[i];
+    const status = section.getAttribute('data-section-status');
```
rofe, line 179: Using the [dataset API](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) with destructuring would be more elegant than the lenghty `getAttribute` imho:
```js
    const { sectionStatus } = section.dataset;
```

scripts/scripts.js
```diff
@@ -157,6 +157,38 @@ export function decorateBlock(block) {
...
+  });
+}
+
+/**
+ * Updates all section status in a container element.
+ * @param {Element} $main The container element
+ */
+function updateSectionsStatus($main) {
+  const sections = [...$main.querySelectorAll(':scope > div.section-wrapper')];
+  for (let i = 0; i < sections.length; i += 1) {
```
rofe, line 177: Why not use `forEach`?

scripts/scripts.js
```diff
@@ -236,12 +268,14 @@ export async function loadBlock(block, eager = false) {
...
  * @param {Element} $main The container element
- * @returns {Array} of Promises to be resolved with blocks are loaded
  */
-export function loadBlocks($main) {
-  const blockPromises = [...$main.querySelectorAll('div.section-wrapper > div > .block')]
-    .map(($block) => loadBlock($block));
-  return blockPromises;
+export async function loadBlocks($main) {
+  const blocks = [...$main.querySelectorAll('div.block')];
+  for (let i = 0; i < blocks.length; i += 1) {
```
rofe, line 274: Why not use `forEach`?

rofe (changes requested):

scripts/scripts.js
kptdobe: Because of `await loadBlock(blocks[i])` in the loop. `await` in the `for(...)` loop will block the process, `forEach` will not block the process, thus you cannot know when `loadBlock`s will be executed.

kptdobe (approved):

scripts/scripts.js
davidnuescheler: i think we had this conversation before, i think the dataset api is great for things that have frequent access and performance is an important aspect. for this use that's not the case, and the additional visibility helps a lot with debugging and also since there is actually CSS triggered off of the status to squash the CLS i don't think it would even be possible.

PR #52, opened 2022-01-23
refactor(rum): use web-vitals library from rum.hlx3.page

### LINKED ISSUE #50: Replace vendored version of web-vitals with Helix-served version
https://github.com/adobe/helix-rum-collector/pull/68 adds the ability to serve the web vitals module from https://helix-rum-collector-serve-through.hlx3.one/.rum/web-vitals/dist/web-vitals.base.js

With a tweak to the default service, we can serve it from `/.rum/web-vitals/dist/web-vitals.base.js` (so that there is no cross-origin loading in the browser).

Once the PR is merged and the VCL has been updated, we can drop the library and update the `import from` line.

trieloff:
fixes #50

- adds a `delayed.js` as container for delayed loading
- adds `sampleRUM('cwv') to `delayed.js
- loads `web-vitals.js` using classic script to work around CORS issues

URL for testing:

- https://unvendor-cwv--helix-project-boilerplate--adobe.hlx3.page/


head.html
```diff
@@ -1,4 +1,4 @@
 <meta name="viewport" content="width=device-width, initial-scale=1"/>
-<script src="/scripts/scripts.js" type="module" crossorigin="use-credentials"></script>
+<script src="/scripts/scripts.js" type="module" crossorigin="anonymous"></script>
```
trieloff, line 2: This is required to avoid the CORS issue that comes with the dynamic import of the web-vitals module. I'm setting `Access-Control-Allow-Origin: *` when serving the RUM script, but this value is only accepted when no credentials get passed.

The only alternative would be to change the CDN config to serve the module from the same host, but as discussed with @stefan-guggisberg this is something we'd like to avoid.

scripts/scripts.js
```diff
@@ -448,6 +449,9 @@ async function loadPage(doc) {
   await loadLazy(doc);
   // eslint-disable-next-line no-use-before-define
   loadDelayed(doc);
+
+  // Core Web Vitals RUM collection
+  sampleRUM('cwv');
```
trieloff, line 453: This has been missing so far, i.e. CWV wasn't ever collected.

trieloff:
@davidnuescheler in my tests, no CWV data was ever collected successfully, so it seems there is still an issue with the way the library is imported (both before and after).

trieloff:
@davidnuescheler the changes we discussed yesterday have been implemented in b0418ba

davidnuescheler (approved):

davidnuescheler:
did the dynamic import not work?

trieloff:
The dynamic import used the `use-credentials` of the outer `script` tag, which conflicted with using an external host name to import. Changing this to `anonymous` would have broken h2 push, so this is what we ended up with.

PR #55, opened 2022-02-01
feat: block-wrapper

davidnuescheler:
https://block-wrapper--helix-project-boilerplate--adobe.hlx3.page/

vs.

https://main--helix-project-boilerplate--adobe.hlx3.page/

i think we often run into the situation, where a block requires a different `max-width` or `background` from the default content or other blocks.
 
in the past, i usually split them off into different sections to achieve that, which i realized recently is not the best solution as it removes the information about the section boundary.
i noticed that we had a concept that we had in the existing DOM that could help with that. 

in the current main branch we have a DOM that looks as follows...

Screen Shot 2022-01-31 at 5 22 29 PM

there is a somewhat underutilized `<div>` below the `.section-wrapper` that we usually use for the `max-width` of default content and blocks in a section.

if we break this up and group default content and individual blocks to something that looks more like this...
Screen Shot 2022-01-31 at 5 25 01 PM
...it allows for a lot more precise control on a per block basis, when it comes to `max-width` or `background`.

this change doesn't really make too much of a difference to any `CSS` rules that i have found but allowed me to implement relatively complex width/background logic with just CSS, not having to resort to sections either through authoring or auto sections for certain blocks...

Screen Shot 2022-01-31 at 5 28 39 PM

... retroactively, i think this would have made the code much simpler in many projects, and would also have kept the `section` metaphor a lot cleaner in places where we want to group blocks and default content, without being forced to sacrifice that grouping because of layout (`width` & `background`) changes.

when pulling together the PR, i realized the `.section-wrapper` class doesn't make any sense in that construct, as the `<div>` really is the `section` div, so i renamed that, which produces a bit of backwards compatibility issue, however it turns out that that one is very easy to fix with a simple search and replace of `section-wrapper` to `section` across a projects code base.

let me know what you think...

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -146,7 +146,7 @@ export function decorateBlock(block) {
  * Decorates all sections in a container element.
  * @param {Element} $main The container element
  */
-function decorateSections($main) {
+export function decorateSections($main) {
   $main.querySelectorAll(':scope > div').forEach((section) => {
     const wrappers = [];
     let defaultContent = false;
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -116,24 +116,6 @@ export function toClassName(name) {
 }
 
 /**
- * Wraps each section in an additional {@code div}.
- * @param {[Element]} $sections The sections
- */
-function wrapSections($sections) {
-  $sections.forEach(($div) => {
-    if ($div.childNodes.length === 0) {
-      // remove empty sections
-      $div.remove();
-    } else if (!$div.id) {
-      const $wrapper = document.createElement('div');
-      $wrapper.className = 'section-wrapper';
-      $div.parentNode.appendChild($wrapper);
-      $wrapper.appendChild($div);
-    }
-  });
-}
-
-/**
  * Decorates a block.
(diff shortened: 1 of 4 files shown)


scripts/scripts.js
```diff
@@ -155,25 +137,38 @@ export function decorateBlock(block) {
...
 }
 
 /**
  * Decorates all sections in a container element.
  * @param {Element} $main The container element
  */
 export function decorateSections($main) {
-  wrapSections($main.querySelectorAll(':scope > div'));
-  $main.querySelectorAll(':scope > div.section-wrapper').forEach((section) => {
+  $main.querySelectorAll(':scope > div').forEach((section) => {
```
rofe, line 150: Should we take this opportunity to get rid of the `$` notation for good?

rofe (approved):

kptdobe (approved):

fkakatie (approved):

scripts/scripts.js
davidnuescheler: let's do that separately... it is used in a lot of places...

PR #56, opened 2022-02-04
chore: safari img loading suppression

davidnuescheler:
https://safari-img-loading--helix-project-boilerplate--adobe.hlx.live/

vs. 

safari (surprisingly, at least to me) still doesn't support `loading="lazy"` and as a side effect of that feels compelled to load every `img` element (even if it is not attached to the DOM yet) as soon as the `src` attribute is set, which means that the fallback `jpg`/`png` images are always loaded. setting the `src` attribute after attaching the `img` element to containing `picture` element seems to fix that.

Screen Shot 2022-02-04 at 11 25 35 AM

Screen Shot 2022-02-04 at 11 26 29 AM

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -340,10 +340,10 @@ export function createOptimizedPicture(src, alt = '', eager = false, breakpoints
       picture.appendChild(source);
     } else {
       const img = document.createElement('img');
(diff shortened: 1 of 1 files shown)


rofe (approved):

trieloff:
iOS 15.4 will enable lazy loading for Safari: https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes#HTML

PR #66, opened 2022-02-18
feat: fragment block

### LINKED ISSUE #65: Fragment block
Block to dynamically include content from another page by specifying an absolute or relative URL:
- `https://main--helix-project-boilerplate--adobe.hlx3.page/fragment`
- `./fragment`

rofe:
Fix #65 

- https://fragment--helix-project-boilerplate--adobe.hlx3.page/

### DIFF
--- a/blocks/fragment/fragment.js
+++ b/blocks/fragment/fragment.js
@@ -10,21 +10,13 @@ export default async function decorate(block) {
   if (resp.ok) {
     const main = document.createElement('main');
     main.innerHTML = await resp.text();
-    const img = main.querySelector('img');
-    if (img) img.setAttribute('loading', 'lazy');
     decorateMain(main);
     await loadBlocks(main);
-    const sections = [...main.children];
     const blockSection = block.closest('.section');
-    sections.forEach((section, i) => {
-      if (!i) {
-        while (sections[0].firstChild) {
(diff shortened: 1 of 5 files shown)


blocks/fragment/fragment.js
```diff
@@ -0,0 +1,30 @@
...
+} from '../../scripts/scripts.js';
+
+export default async function decorate(block) {
+  const ref = block.textContent.trim();
+  const path = new URL(ref, window.location.href).pathname.split('.')[0];
+  const resp = await fetch(`${path}.plain.html`);
+  if (resp.ok) {
+    const main = document.createElement('main');
+    main.innerHTML = await resp.text();
+    const img = main.querySelector('img');
```
kptdobe, line 13: Why only first image is lazy loaded ?

kptdobe (approved):

kptdobe:
We should probably test (and add a check) for recursive fragment includes...

blocks/fragment/fragment.js
rofe: Because the rest is set to`lazy` anyway. But to be honest, at this point it's irrelevant anyway, so I'll remove it again.

rofe:
It's still unclear what should happen with multiple sections, so I decided to remove multi-section support for now. We can add it back when we have a concrete use case.

PR #67, opened 2022-02-21
chore: remove smart sizing and colors

davidnuescheler:
https://unclever--helix-project-boilerplate--adobe.hlx3.page/

vs.

https://main--helix-project-boilerplate--adobe.hlx3.page/

there were a number of clever aspects of the design that don't map to what's typically encountered in projects (eg. font sizes in `rem` and `%` or colors as decimal rgb values) so this PR removes all of that and replaces it with a more typical scenario encountered in projects.

### DIFF
--- a/blocks/header/header.css
+++ b/blocks/header/header.css
@@ -32,7 +32,7 @@ header .nav[aria-expanded='false'] .nav-hamburger-icon::before {
   width: 20px;
   height: 2px;
   border-radius: 3px;
-  background: currentColor;
+  background: currentcolor;
 }
 
 header .nav[aria-expanded='false'] .nav-hamburger-icon::after,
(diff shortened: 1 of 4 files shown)


kptdobe (approved):

blocks/header/header.css
```diff
@@ -32,7 +32,7 @@ header .nav[aria-expanded='false'] .nav-hamburger-icon::before {
   width: 20px;
   height: 2px;
   border-radius: 3px;
-  background: currentColor;
+  background: currentcolor;
```
rofe, line 35: Isn't it `currentColor`? https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword

rofe (approved):

blocks/header/header.css
davidnuescheler: i also thought that, but the css linter complained. let's make sure the linter is configured correctly...

fkakatie (approved):

blocks/header/header.css
rofe: According to https://stylelint.io/user-guide/rules/list/value-keyword-case/ all keyword values are expected to be lowercase, referring to https://www.w3.org/TR/css-color-4/#currentcolor-color.

I'd say we stand corrected 😄

PR #81, opened 2022-04-12
chore: add json to codebus

davidnuescheler:
not sure why we excluded all `.json`, maybe because of the `.json`s in the root?


tripodsan:
> not sure why we excluded all `.json`, maybe because of the `.json`s in the root?

yes. I added an exclusion for them

tripodsan (approved):

PR #82, opened 2022-04-25
chore: initial nav refactor

davidnuescheler:
this is a refactor that changes the content structure of the nav to something that seems more inline with the DOM that is semantically correct....

https://nav-refactor--helix-project-boilerplate--adobe.hlx.page/nav-new.plain.html

vs.

https://main--helix-project-boilerplate--adobe.hlx.page/nav.plain.html

to illustrate a very common tools section i added and search icon and also implemented the `icon` handling that embeds the SVG into the DOM.

there are a few other aspects with respect to the color setup to further unclever things... i didn't spend too much time thinking about it, as know that @fkakatie will probably refactor this anyway.


rofe (changes requested):
Can you fix the `stylelint` issues?

rofe (approved):

blocks/header/header.js
```diff
@@ -21,49 +21,47 @@ export default async function decorate(block) {
   block.textContent = '';
 
   // fetch nav content
-  const navPath = cfg.nav || '/nav';
+  const navPath = cfg.nav || '/nav-new';
```
rofe, line 24: I assume we would revert that before merging?

blocks/header/header.js
```diff
@@ -21,49 +21,47 @@ export default async function decorate(block) {
...
   const resp = await fetch(`${navPath}.plain.html`);
-  const html = await resp.text();
+  if (resp.status === 200) {
```
rofe, line 26:
```suggestion
  if (resp.ok) {
```
Might also include `304`

rofe (approved):

kptdobe (approved):

PR #85, opened 2022-04-28
Icon handling

davidnuescheler:
this is a client sided implementation of https://github.com/adobe/helix-pipeline-service/issues/386
to see if it is feasible.

https://icon-handling--helix-project-boilerplate--adobe.hlx.page/

as there are currently no icons in the project there is not much to see here.

### DIFF
--- a/blocks/header/header.js
+++ b/blocks/header/header.js
@@ -1,4 +1,4 @@
-import { readBlockConfig } from '../../scripts/scripts.js';
+import { readBlockConfig, decorateIcons } from '../../scripts/scripts.js';
 
 /**
  * collapses all open nav sections
@@ -64,6 +64,6 @@ export default async function decorate(block) {
   });
   nav.prepend(hamburger);
   nav.setAttribute('aria-expanded', 'false');
-
+  await decorateIcons(nav);
   block.append(nav);
 }
--- a/blocks/header/header.js
+++ b/blocks/header/header.js
@@ -64,6 +64,6 @@ export default async function decorate(block) {
   });
   nav.prepend(hamburger);
   nav.setAttribute('aria-expanded', 'false');
-  await decorateIcons(nav);
+  decorateIcons(nav);
   block.append(nav);
 }
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -140,13 +140,15 @@ export function decorateIcons(element) {
   // prepare for forward compatible icon handling
   replaceIcons(element);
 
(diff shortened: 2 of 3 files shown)


davidnuescheler:
@kptdobe i am sure that the `await` in the `.forEach` loop could be improved, and i am sure that you might have the best solution handy.

davidnuescheler:
i think this is easiest to see in action on here https://nav-refactor--helix-project-boilerplate--adobe.hlx.page/nav-new as this includes an icon in the nav that is used in different places with different CSS context.

scripts/scripts.js
```diff
@@ -120,6 +120,36 @@ export function toClassName(name) {
...
+    span.className = img.className;
+    img.replaceWith(span);
+  });
+}
+
+/**
+ * Replace icons with inline SVG and prefix with codeBasePath.
+ * @param {Element} element
+ */
+export async function decorateIcons(element) {
```
kptdobe, line 139:
```suggestion
export function decorateIcons(element) {
```

As is, the method is not `async`.

blocks/header/header.js
```diff
@@ -64,6 +64,6 @@ export default async function decorate(block) {
   });
   nav.prepend(hamburger);
   nav.setAttribute('aria-expanded', 'false');
-
+  await decorateIcons(nav);
```
kptdobe, line 67:
```suggestion
  decorateIcons(nav);
```

As is, the method is not `async`.

kptdobe:
> @kptdobe i am sure that the await in the .forEach loop could be improved, and i am sure that you might have the best solution handy.

It depends. I think it is fine as is because we do not care of the order and the promises will be executed by the browser (not risk to exit before they are all executed). This should just be reflected in the `decorateIcons` method signature: it is not async.
If you really want to make it async, then yes, you need to change the loop to something like (not tested):

```
export async function decorateIcons(element) {
  replaceIcons(element);

  const promises = [];
  element.querySelectorAll('span.icon').forEach((span) => {
    promises.push(new Promise((resolve) => {
      const iconName = span.className.split('icon-')[1];
      const resp = await fetch(`${window.hlx.codeBasePath}/icons/${iconName}.svg`);
      if (resp.status === 200) {
        const svg = await resp.text();
        span.innerHTML = svg;
      }
      resolve();
    });
  });

  await Promise.all(promises);
}
```

scripts/scripts.js
rofe: What about the `await fetch()`?

PR #99, opened 2022-05-24
feat: custom 404 example

### LINKED ISSUE #63: Custom 404 example
Add example 404 handling.

fkakatie:
add sample 404 page

- ~~attempting to balance simplicity with visual interest (my hope is that the "0" is somewhat reminiscent of a faulty helix strand)~~
- a "go back" button will be added in addition to the "go home" button if there is a `document.referrer` and its origin matches the `window.location`'s origin

Fix #63 

Test URLs:
- Before: https://main--helix-project-boilerplate--adobe.hlx.page/error
- After: https://404--helix-project-boilerplate--adobe.hlx.page/error

### DIFF
--- a/404.html
+++ b/404.html
@@ -23,7 +23,7 @@
           backBtn.textContent = 'Go back';
           backBtn.title = 'Go back';
           const btnContainer = document.querySelector('.button-container');
-          btnContainer.prepend(backBtn);
+          btnContainer.append(backBtn);
         }
       }
     });
--- a/404.html
+++ b/404.html
@@ -120,7 +120,7 @@
       </p>
       <h2 class="error-message">Page Not Found</h2>
       <p class="button-container">
-        <a href="/" class="button secondary">Go home</a></p>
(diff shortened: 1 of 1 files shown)


kptdobe (changes requested):
Independently of liking the design or not, I think it is important to build the page in a way that nothing has to be done when creating a new project (no project name to update, no url to adjust...).
Looking at this nice page with the rotating Helix, I somehow changed my mind and tend to agree with @davidnuescheler regarding extra simplicity: every project will have its own custom 404 page... at one point. Building that page will certainly not be the priority, the default one will be there for a while and even potentially go-live initially. It should then be a nice clean page that would be an acceptable default for every project.

rofe (approved):
Too bad I missed the version with the rotating helix! 😄

trieloff (approved):

kptdobe (approved):

davidnuescheler (approved):

PR #103, opened 2022-05-30
feat: cards

davidnuescheler:
https://cards--helix-project-boilerplate--adobe.hlx.live/drafts/uncled/

### DIFF
--- a/blocks/cards/cards.css
+++ b/blocks/cards/cards.css
@@ -12,15 +12,15 @@
     background-color: var(--background-color)
 }
 
-.cards .cards-body {
+.cards .cards-card-body {
     margin: 16px;
 }
 
-.cards .cards-image {
+.cards .cards-card-image {
     line-height: 0;
 }
 
-.cards .cards-body *:first-child {
+.cards .cards-card-body > *:first-child {
     margin-top: 0;
 }
 
--- a/blocks/cards/cards.css
+++ b/blocks/cards/cards.css
@@ -1,32 +1,31 @@
 .cards > ul {
-    display: grid;
-    grid-template-columns: repeat(auto-fill, minmax(15rem, 1fr));
-    grid-gap: 16px;
     list-style: none;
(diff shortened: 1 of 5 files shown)


blocks/cards/cards.css
```diff
@@ -0,0 +1,32 @@
...
+    border: 1px solid var(--overlay-background-color);
+    border-radius: 10px;
+    transition: all 0.5 ease-out; 
+    box-shadow: 1px 1px 5px #00000020;
+}
+
+.cards > ul > li:hover {
+    box-shadow: 2px 2px 10px #00000040;
+    transform: translate(-1px, -1px)
+}
```
fkakatie, line 21: I think is is potentially confusing... this makes me think that the card is clickable.

blocks/cards/cards.js
```diff
@@ -0,0 +1,19 @@
...
+      cell.className = classNames[i] ? `${blockName}-${classNames[i]}` : '';
+    });
+  };
+
+  /* change to ul, li */
+  const ul = document.createElement('ul');
+  [...block.children].forEach((row) => {
+    const li = document.createElement('li');
+    li.innerHTML = row.innerHTML;
+    applyClasses([...li.children], ['image', 'body']);
```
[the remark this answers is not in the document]
davidnuescheler: that's great. i think you are correct, i cleaned it up a little bit and also changed the logic as i realized that there are quite afew card usecases that i found that needed to be flexible in terms of dealing with card that either have an image or not...

as for the namespacing (starting all the classes in a block with the blockname as the prefix) i left that in, but i feel that i am violating the the intuitive -> simple -> consistent rule here. hmm...

blocks/cards/cards.css
davidnuescheler: true true... fixed

fkakatie (approved):

rofe (approved):

PR #109, opened 2022-06-20
Fix List Enumeration in README.md

LucaNerlich:
Tiny fix for inconsistent README list enumeration.


trieloff:
This is actually intentional. Ordered lists in Markdown will be turned into sequentially ordered lists in HTML, even if the input is just random numbers, so keeping every entry at `1.` will make it easier to add new entries in the middle of the list.

LucaNerlich:
Alright, that makes sense to me. Thanks :)

Ive never ran into that issue, since most Editors fix List Enumeration automatically, when inserting new entries.

PR #135, opened 2022-10-18
[WIP] feat: minimal plugin system PoC

### LINKED ISSUE #137: Minimal plugin system
## Use case

As a developer, I'd like to have a minimal plugin system so that I can organize common features into individual opt-in plugins that I can re-use across projects, so that my `scripts.js` is easier to maintain, my code more portable and re-usable, and so that I only load what is really needed for my use case.

In particular, I'd like to:
- have a simple instrumentation of my `scripts.js` for the plugins I use
- have the capability to pass down configuration settings to the plugin to tailor it for my project
- be able to define various "hooks" in my plugin so it automatically runs in the eager/lazy/delayed phases
- be able to reference another plugin from my plugin so I can re-use features

ramboz:
## Use case

As a developer, I'd like to have a minimal plugin system so that I can organize common features into individual opt-in plugins that I can re-use across projects.

## Driving principles

- Protection of the loading sequence (Eager, Lazy, Delayed)
- Separation of concerns
- Increase code testability
- Keep PIS score at 100

## Technical

- plugins can be loaded via the `await withPlugin(<path>, <options>)` method
- protect the loading phases by moving the logic inside `lib-franklin.js`
- Offer `pre`/`post` hooks for the `Eager`/`Lazy`/`Delayed` phases so plugins can inject logic at the most appropriate moment in the loading lifecycle
- `pre`/`post` hooks get the `options` object as well as a map of plugins that expose an API. For instance
    ```js
    // this isn't publicly exposed and only available to direct importers
    export function aMethodNotExposedInTheApi(…) { … }

    // this is publicly exposed and passed down in the pre/post hooks
    export const API = {
      aMethodExposedInTheApi: (…) => { … } 
    }

    // defining a running before the Eager phase
    export async function preEager(options, plugins) {
      // options.myOption
      // plugins.myPlugin.myMethod(…)
    }
    ```
- Plugins can expose a limited API usable via `const myApi = await withPlugin('./plugins/myPlugin.js', …);`
- Simple eager/lazy/delayed hooks in the project via the usual `load*` methods and a configurable delayed timeout:
    ```js
    async function loadEager(doc, options) { … }
    async function loadLazy(doc, options) { … }
    async function loadDelayed(doc, options) { … }

    init({
      loadEager,
      loadLazy,
      loadDelayed,
      delayedDuration: 3000
    });
    ```
- RUM,  decoration and placeholders logic extracted as plugins
- Leveraging `<link rel="modulepreload">` for the plugins to speed up page load
- All plugin methods are executed with a context that gives access to the main lib-franklin utils (via `this`)

The loading sequence will be:
- `init` of all plugins
- `preEager` of all plugins
- `loadEager` for the project
- `postEager` of all plugins
- `preLazy` of all plugins
- `loadLazy` for the project
- `postLazy` of all plugins
- … _configurable timeout_ … _3s_
- `preDelayed` of all plugins
- `loadDelayed` for the project
- `postDelayed` of all plugins

## Performance comparison with `main`

Screen Shot 2022-10-26 at 11 38 27 AM
Screen Shot 2022-10-26 at 11 37 04 AM

Fix #137

Test URLs:
- After: https://poc-plugin-system--helix-project-boilerplate--adobe.hlx.live/


scripts/scripts.js
```diff
@@ -8,13 +7,17 @@ import {
...
   loadBlocks,
   loadCSS,
+  waitForLCP,
+  withPlugin,
 } from './lib-franklin.js';
 
 const LCP_BLOCKS = []; // add your LCP blocks to the list
-window.hlx.RUM_GENERATION = 'project-1'; // add your RUM generation information here
+
+const rum = await withPlugin('./plugins/rum.js');
```
kptdobe, line 19:
```suggestion
const rum = await withPlugin('./plugins/rum.js', 'project-1');
```
to make clear that you have to touch here and not in `rum.js`.

head.html
```diff
@@ -1,4 +1,5 @@
 <meta name="viewport" content="width=device-width, initial-scale=1"/>
+<link rel="modulepreload" href="/scripts/plugins/rum.js"/>
```
kptdobe, line 2: Really necessary ?

ramboz: To some extent, it helps with the loading. Since plugins are loaded async by `scripts.js`, this inherently introduces a delay before the lazy phase kicks in. So preloading the key plugins as part of the initial 6 parallel requests we are allowed speeds it up a notch.

Before:
Screen Shot 2022-10-19 at 12 11 18 PM

After:
Screen Shot 2022-10-19 at 12 11 30 PM

ramboz:
Some other ideas to explore:

1. Expose a chainable object, instead of individual methods:
    ```js
    Franklin
      .withPlugin('./plugins/foo.js', { … })
      .withPlugin('./plugins/bar.js', { … })
      .init({
        loadEager,
        loadLazy,
        loadDelayed,
      });
    ```
2. Add option to make some plugins "blocking" if they are needed in the eager phase, versus non-blocking (by default) if they are only needed in the lazy phase:
    ```js
    await withPlugin('./plugins/foo.js', { … });
    withPlugin('./plugins/bar.js', { … });
    ```
    or maybe via options `{ isBlocking: true }` or similar
3. Add plugins pre-loading via `Link` headers

scripts/plugins/rum.js
```diff
@@ -0,0 +1,86 @@
...
+    // something went wrong
+  }
+}
+
+export const api = {
+  sampleRUM,
+};
+
+export async function preEager(options) {
+  window.hlx.RUM_GENERATION = options.projectName; // add your RUM generation information here
```
trieloff, line 70: I'd keep the setting of `window.hlx.RUM_GENERATION` out of the plugin (so there is less incentive to change the plugin).

Using the `projectName` is wrong, because the generation should describe the version used for RUM collection or the particular code approach, so that you can compare the performance of two generations against each other.

trieloff:
```suggestion
  window.hlx.RUM_GENERATION = options.generation; // add your RUM generation information here
```

scripts/scripts.js
```diff
@@ -1,20 +1,20 @@
...
 const LCP_BLOCKS = []; // add your LCP blocks to the list
-window.hlx.RUM_GENERATION = 'project-1'; // add your RUM generation information here
+
+const {
+  decorateBlock,
+  decorateButtons,
+  decorateIcons,
+  decorateTemplateAndTheme,
+} = await withPlugin('./plugins/decorator.js');
+const rum = await withPlugin('./plugins/rum.js', { projectName: 'project-1' });
```
trieloff, line 17:
```suggestion
const rum = await withPlugin('./plugins/rum.js', { generation: 'project-1' });
```

scripts/plugins/rum.js
```diff
@@ -0,0 +1,86 @@
...
+        navigator.sendBeacon(url, body);
+        // eslint-disable-next-line no-console
+        console.debug(`ping:${checkpoint}`, pdata);
+      };
+      sampleRUM.cases = sampleRUM.cases || {
+        cwv: () => sampleRUM.cwv(data) || true,
+        lazy: () => {
+          // use classic script to avoid CORS issues
+          const script = document.createElement('script');
+          script.src = 'https://rum.hlx.page/.rum/@adobe/helix-rum-enhancer@^1/src/index.js';
```
trieloff, line 51: The code loaded behind the scenes assumes that `sampleRUM` is a global object. Your changes break that contract, rendering the advanced functionality useless.

test/scripts/block-utils.test.js
```diff
@@ -104,20 +85,20 @@ describe('Sections and blocks', () => {
 
   it('Updates section status', async () => {
     blockUtils.updateSectionsStatus(document.querySelector('main'));
-    document.querySelectorAll('main .section').forEach(($section) => {
+    document.querySelectorAll('main > div').forEach(($section) => {
```
trieloff, line 88: Are these test changes indicative of breaking changes?

trieloff (changes requested):
I think this breaks the advanced RUM functionality (Intersection Observer)

test/scripts/block-utils.test.js
ramboz: Not that I could identify so far. I was trying to decouple the loading status logic from the block/section decoration… but it's too tightly linked so will likely revert and make the decorator plugin "mandatory"

scripts/plugins/rum.js
ramboz: @trieloff You mean that it is assuming it's exposed on `window.hlx.rum`, right? This is still the case in the plugin. The instrumentation is just isolated in the plugin and decoupled from `scripts.js`, but I haven't changed existing APIs as far as I can tell.

trieloff: Screen Shot 2022-10-24 at 15 52 08

The `load` checkpoint is never sent from your refactored script. See #139 (which will conflict) for the correct checkpoint sequence.

ramboz: @trieloff Thanks for sharing, that helps set clear expectations. Are there clear expectations for each event? Do we have those documented somewhere?
For instance, does the load event need to be triggered on the page load? or is it meant to reflect a specific moment in the loading phase?

ramboz: There are various ways I can fix the missing `load` event depending on the exact expectation.

Since the RUM plugin is loaded async, essentially the load event would have fired by the time any code in it is executed. So, in theory, I can bring it back into the main script and lose the separation of concern if we care about the exact page load event.

But I could also just fire it whenever the rum plugin has loaded, if what we care is when the main JS logic is loaded.

Or I could defer to when the content becomes visible (waitForLCP), if that is what we really care about.

It really depends what we use that specific event for in the end in the analytics reports

ramboz: For instance https://github.com/adobe/helix-project-boilerplate/pull/135/commits/0d2bab2646aaf325699e10707f36c6034da3e8cc adds the event back, and we essentially now have:
- `top` is fired when the plugin is "parsed"
- `load` is fired when the plugin is "initialized"

trieloff: `top` should be fired as soon as possible. The goal is to catch any hint of the page being interacted with at all.
`load` is the https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event plain and simple.

ramboz: @trieloff I moved the rum plugin back into lib-franklin so we can trigger on the DOM load event

ramboz:
@trieloff, @kptdobe I think I addressed all the concerns you had. Do you see anything else that would need some love here?

I don't want to spend too much time working on bumping test coverage if you still have concerns with the underlying proposal.

trieloff:
Why don't we try this in a project and see if the benefits of maintainability pan out?

ramboz:
@meryllblanchet Maybe something to test with one of your projects that are in the pipeline?

ramboz:
Closing in favour of a more lightweight version in #254

PR #149, opened 2022-12-02
feat: auto-sprite svg icons

ramboz:
## Context
See https://github.com/BambooHR/bamboohr-website/pull/75 for full context

## Problem
On pages that massively use icons, the repeated loading of those via `fetch` requests delays the rendering of the lazy parts of the page (header, footer, etc.)

## The solution
We create an inline SVG sprite in the document and dynamically populate it when we encounter new icons.

Screenshot 2022-12-02 at 9 47 08 AM

On the boilerplate that doesn't really use any icons, the change is essentially transparent.

Fix https://github.com/BambooHR/bamboohr-website/pull/75

Test URLs:
- After: https://inline-svg-sprite--helix-project-boilerplate--ramboz.hlx.live/

### DIFF
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -139,27 +139,34 @@ export async function decorateIcons(element) {
 
   const icons = [...element.querySelectorAll('span.icon')];
 
-  icons.forEach(async (span) => {
+  return Promise.all(icons.map(async (span) => {
     const iconName = span.classList[1].substring(5);
+    let html;
     if (!symbols[iconName]) {
       symbols[iconName] = true;
       try {
         const response = await fetch(`${window.hlx.codeBasePath}/icons/${iconName}.svg`);
         const svg = await response.text();
-        svgSprite.innerHTML += svg
-          .replace('<svg', `<symbol id="${iconName}"`)
-          .replace('</svg>', '</symbol>');
+        // Styled svg icons don't play nice with the `<use>` references, so just inline those
+        if (svg.match(/(<style | class=)/)) {
+          html = svg;
+        } else {
+          // Add the icon to the sprite without its hardcoded size so we can style it in css
+          svgSprite.innerHTML += svg
+            .replace('<svg', `<symbol id="${iconName}"`)
+            .replace(/ width=".*?"/, '')
+            .replace(/ height=".*?"/, '')
+            .replace('</svg>', '</symbol>');
+          html = `<svg xmlns="http://www.w3.org/2000/svg"><use href="#${iconName}"/></svg>`;
+        }
       } catch (err) {
         console.error(err);
       }
     }
 
     const parent = span.firstElementChild?.tagName === 'A' ? span.firstElementChild : span;
-    parent.innerHTML = `
-      <svg xmlns="http://www.w3.org/2000/svg">
-        <use href="#${iconName}"/>
-      </svg>`;
-  });
+    parent.innerHTML = html;
+  }));
 }
 
 /**
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -123,12 +123,13 @@ export function toCamelCase(name) {
   return toClassName(name).replace(/-([a-z])/g, (g) => g[1].toUpperCase());
 }
 
+const ICONS_CACHE = {};
 /**
  * Replace icons with inline SVG and prefix with codeBasePath.
  * @param {Element} element
  */
 export async function decorateIcons(element) {
-  const symbols = {};
+  // Prepare the inline sprite
   let svgSprite = document.getElementById('franklin-svg-sprite');
   if (!svgSprite) {
     const div = document.createElement('div');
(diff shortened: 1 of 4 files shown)


rofe (approved):
I like the approach!

scripts/lib-franklin.js
```diff
@@ -127,24 +127,38 @@ export function toCamelCase(name) {
...
+  if (!svgSprite) {
+    const div = document.createElement('div');
+    div.innerHTML = '<svg xmlns="http://www.w3.org/2000/svg" id="franklin-svg-sprite" style="display: none"></svg>';
+    svgSprite = div.firstElementChild;
+    document.body.append(div.firstElementChild);
+  }
+
+  const icons = [...element.querySelectorAll('span.icon')];
+
+  icons.forEach(async (span) => {
```
trieloff, line 142: You might be better served by a `await Promise.all(icons.map(…)))` so that a single blocking sprite won't block the remaining icons.

kptdobe (changes requested):

ramboz:
PSI check seems flaky, but deep PSI doesn't see a performance impact… which for a single icon in the lazy phase isn't necessarily that reliable anyway
Screenshot 2023-01-31 at 10 39 02 AM

ramboz:
@trieloff @kptdobe @davidnuescheler Took some time to revive this and address open comments. Please have a look when you have some time.

kptdobe:
> which for a single icon in the lazy phase isn't necessarily that reliable anyway

In the initial Bamboohr issue, you explicitly say `pages that use a lot of icons`.
I am not convinced that the PR addresses the problem because I believe the sprite is just a different implementation, not the source of the problem. For me resides in the `fetch` being called multiple times for the same icon (because of the `Promise.all`, I think we still go through identical requests) or N times if you have N icons (if N is "large", you will definitively still see a performance impact). 
I might be wrong thus that's probably 2 test cases to measure (could be done on the BambooHR page) before deciding if this PR addresses the issue or not.

ramboz:
@kptdobe You are correct. This will only avoid repeatedly downloading the _same_ icon 10s or 100s of times on a page that repeatedly uses it. But if you have 100 different icons, we'll still do 100 fetch calls. So far, I've only encountered the "table-like" structure like BambooHR had as a real problem, so that's what I optimized for.

During my tests on BambooHR, I had tried to compare the "100 different icons" vs. "1 SVG sprite with 100 icons". The former was visually more appealing for the end user as icons in the viewport load faster, as opposed to having to load the whole sprite. And from a performance point of view, this only marginally influenced TTI and SI scores.

So essentially, you won't see any real change on regular websites, and just see a TTI/SI boost on pages that massively use the same icons over and over again.

If you have any ideas to address the 100 different icons use case, I'm happy to do some explorations. But so far, just chaining the calls seemed the most appropriate in the end

scripts/lib-franklin.js
```diff
@@ -127,25 +127,46 @@ export function toCamelCase(name) {
...
+    document.body.append(div.firstElementChild);
+  }
+
+  const icons = [...element.querySelectorAll('span.icon')];
+
+  return Promise.all(icons.map(async (span) => {
+    const iconName = span.classList[1].substring(5);
+    let html;
+    if (!symbols[iconName]) {
+      symbols[iconName] = true;
```
hannessolo, line 146: I tried using this code in https://github.com/hlxsites/maidenform/issues/114, but I think your latest commit broke the loading here.

If `symbols[iconName]` is `true`, `html` will be undefined.

In my code, I addressed this by saving the `html` variable to the map `symbols[iconName]`, instead of storing a boolean, and setting the parent `innerHTML` to this.

scripts/lib-franklin.js
```diff
@@ -123,28 +123,65 @@ export function toCamelCase(name) {
...
+    const div = document.createElement('div');
+    div.innerHTML = '<svg xmlns="http://www.w3.org/2000/svg" id="franklin-svg-sprite" style="display: none"></svg>';
+    svgSprite = div.firstElementChild;
+    document.body.append(div.firstElementChild);
+  }
+
+  // Download all new icons
+  const icons = [...element.querySelectorAll('span.icon')];
+  await Promise.all(icons.map(async (span) => {
+    const iconName = span.className.split('icon-')[1];
```
hannessolo, line 144: Compared to the previous code, this will break for cases where icons have extra class names, like: `icon icon-hamburger blue` -> `iconName` becomes `icon-hamburger blue`.

hannessolo (changes requested):

scripts/lib-franklin.js
```diff
@@ -123,28 +123,65 @@ export function toCamelCase(name) {
...
+        ICONS_CACHE[iconName] = false;
+        console.error(err);
       }
     }
+  }));
+
+  const symbols = Object.values(ICONS_CACHE).filter((v) => !v.styled).map((v) => v.html).join('\n');
+  svgSprite.innerHTML += symbols;
+
+  icons.forEach(async (span) => {
```
hannessolo, line 176: I think this function doesn't need to be async anymore?

davidnuescheler:
i think that we should prefix the id= with something like `icons-sprite-${iconName}`.
possibly the `iconName` should be created from a `.substring(5)` not from a `.split('-')[1]` and i think there is a redundant `${window.hlx.codeBasePath}` in the `fetch()`

kptdobe (approved):

PR #155, opened 2022-12-15
feat: better nav override

davidnuescheler:
introducing a page level metadata override of nav...

https://header--helix-project-boilerplate--adobe.hlx.page/drafts/uncled/

vs. 

https://main--helix-project-boilerplate--adobe.hlx.page/drafts/uncled/


blocks/header/header.js
```diff
@@ -17,11 +17,11 @@ function collapseAllNavSections(sections) {
  */
 
 export default async function decorate(block) {
-  const cfg = readBlockConfig(block);
   block.textContent = '';
 
   // fetch nav content
-  const navPath = cfg.nav || '/nav';
+  const navMeta = getMetadata('nav');
+  const navPath = navMeta ? new URL(navMeta).pathname : '/nav';
```
nc-andreashaller, line 24: I would recommend to wrap this line in a function with a try-catch to avoid breaking the header loading completely if a configuration is broken.

rofe (approved):

fkakatie (approved):

blocks/header/header.js
davidnuescheler: i think that's a good point in case someone for example added something that's not a well-formed URL.

having said that, i would probably want the author (who is very likely a developer, in this case) to see that the nav is broken visually as opposed to falling back to a default nav. considering that this line is probably replaced with something that includes some locale handling, i think i will keep it simple at this juncture.

fkakatie (approved):

PR #158, opened 2023-01-04
Accessible header

fkakatie:
increase accessibility of header block by:
- wrapping `.nav-hamburger-icon` in `button` (semantic element, native keyboard behavior and implicit role)
- adding `aria-label` attribute as the button otherwise has no text description
- making relationship between hamburger button and nav more explicit (linked by `aria-controls` as the button and the menu are not adjacent siblings)
- updating `aria-expanded` values for the entire nav as well as nested nav sections on viewport size change
  - for example, changing the viewport width while the hamburger menu is open no longer leaves strange artifact
Screen Shot 2023-01-03 at 7 14 24 PM

- expanding keyboard navigation for nav and nav sections
  - in mobile/tablet view, hamburger menu can be opened and closed with the space or enter keys while focusing on the hamburger button (non-links are excluded from keyboard navigation)
  - in desktop view, the nav sections can be opened and closed with the space or enter keys while focusing on the controlling `li` (appropriate `role` and `tabindex` have been added)
  - both the hamburger menu and the nav sections can be closed by using the escape key, then focus is redirected to the element that controls the element hidden by the collapse (an attempt not to "lose" the focus when a menu is collapsed)
- maintaining the height of the nav when the hamburger menu is opened so the control always appears in the same place (and adding more clear visual distinction between nav section descriptions and actual links)
Screen Shot 2023-01-03 at 7 14 24 PM

related changes:
- removed or corrected invalid/incomplete css properties (box-shadow, currentolor [sic])
- unified padding between main content and header
- centered the search svg on the artboard
- changed the `body-font-size-xs` size for use in header (this variable is not used elsewhere in the project)

Test URLs:

- After: https://accessible-header--helix-project-boilerplate--adobe.hlx.page/
- Diffsite: https://pianomister.github.io/diffsite/?url1=https%3A%2F%2Fmain--helix-project-boilerplate--adobe.hlx.page%2F&url2=https%3A%2F%2Faccessible-header--helix-project-boilerplate--adobe.hlx.page%2F


davidnuescheler (approved):
amazing! 🎆

blocks/header/header.css
```diff
@@ -1,260 +1,273 @@
...
-  padding: 0 2rem;
+  gap: 0 2em;
   position: fixed;
+  left: 50%;
+  transform: translateX(-50%);
   z-index: 1;
-  background-color: var(--background-color);
-  width: 100vw;
+  width: 100%;
+  max-width: 1200px;
```
rofe, line 15: Align with content max-width
```suggestion
  max-width: 1264px;
```

rofe (approved):

blocks/header/header.css
fkakatie: good catch! thank you!

PR #162, opened 2023-01-10
fix: vertical scrollbar disappears on desktop

### LINKED ISSUE #161: Cannot scroll the content
Open https://main--helix-project-boilerplate--adobe.hlx.page/
The scrollbar disappears after a few ms and the content cannot be scrolled.

An `overflow-y: hidden` attribute is added to the body. I suspect the new header to have introduced this regression (https://github.com/adobe/helix-project-boilerplate/blob/main/blocks/header/header.js#L58 ?) 
@fkakatie can you please have a look ?

fkakatie:
currently, scrollbar will disappear on desktop sizes when the nav is set to "expanded"
now, scrollbar will disappear only when nav is "expanded" on mobile/tablet sizes (previous behavior) 

Fix #161 

Test URLs:
- After: https://header-fix--helix-project-boilerplate--adobe.hlx.page/


dylandepass (approved):
🚀

PR #164, opened 2023-01-11
feat: image-only columns will display first on mobile view

fkakatie:
when a column in the columns block is an image-only column, it will display before (or above) other column row content.

Test URLs:
- After: https://columns--helix-project-boilerplate--adobe.hlx.page/

### DIFF
--- a/blocks/columns/columns.css
+++ b/blocks/columns/columns.css
@@ -1,25 +1,33 @@
 .columns > div {
-    display: flex;
-    flex-direction: column;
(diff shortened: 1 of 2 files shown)


fkakatie:
I personally have never seen any site where images above text on mobile view is not observed, whether or not there is a z-pattern/river flow/whatever, but does the extra JS merit inclusion in block collection rather than boilerplate?

davidnuescheler (approved):

PR #177, opened 2023-02-11
chore: better nav reload after publish

davidnuescheler:
there are some repeated questions related to browser cache after publishing changes to `nav` or `footer`.

here is some simple logic that forces a `reload` of the `.plain.html`

https://main--helix-project-boilerplate--adobe.hlx.live/nav
vs.
https://nav-reload--helix-project-boilerplate--adobe.hlx.live/nav


fkakatie (approved):
wow, that was so much easier than I envisioned.

PR #178, opened 2023-02-12
Minor Nav touch-up ideas

davidnuescheler:
couple of ideas for the nav

https://nav-touchups--helix-project-boilerplate--adobe.hlx.live/
vs.

(1) introducing nav wrapper:

Screenshot 2023-02-12 at 3 51 09 PM
Screenshot 2023-02-12 at 3 51 34 PM

(2) minor naming proposal:
`MQ` seems a bit generic, i think semantically it would be `isDesktop`

(3) nav config via meta:
somehow we seem to have used block config to specify the nav path, but i don't think that's practical way of doing it, i would propose to use meta instead

(4) bumping z-index:
i think hero (and any other background image situation) should use `z-index` `0` and `1` instead of the current `-1` setup, since that one is not compatible with simple background colors.

### DIFF
--- a/blocks/header/header.css
+++ b/blocks/header/header.css
@@ -1,4 +1,11 @@
 /* header and nav layout */
+header .nav-wrapper {
+  background-color: var(--background-color);
+  width: 100%;
+  z-index: 2;
+  position: fixed;
+}
+
 header nav {
   box-sizing: border-box;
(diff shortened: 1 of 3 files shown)


fkakatie (approved):
🎉

PR #182, opened 2023-02-24
Dynamic Template Support

shsteimer:
Opening this as a draft PR to trigger discussion.

This version of templating was implemented for Keysight, and has been asked about on slack https://cq-dev.slack.com/archives/C9KD0TT6G/p1677182933672489 and here https://cq-dev.slack.com/archives/C9KD0TT6G/p1674524170715789 so I feel like that is at least enough interest to consider if it should go in boilerplate.

You can see real pages using this at:
After: https://main--blogs-keysight--hlxsites.hlx.live/authors/rakesh-seal
After: https://main--blogs-keysight--hlxsites.hlx.live/tech/nwvs/2023/01/31/krpc-protocol-the-language-of-torrent-peers

To attempt to address some points I expect to come up:
* Is this something every customer needs, to the point it should be in the boilerplate?

Not exactly. I think it's common enough that it needs to go somewhere that is easily reference-able/usable when it is needed, but the number is probably somewhere around 30-50% of customers, if I had to guess. 

If it goes in boilerplate, then it needs to be done in a way that customers who don't need it don't need to delete anything. I think doing it this way (the code to support template css/js is there, but no actual templates are implemented, and are left to individual projects to add if needed) meets that requirement. We'd probably also want an example of a template implementation that is not the boilerplate. WKND would be the most likely home for that, so I can see the case for just putting the whole thing there and leaving the boilerplate clean. Ultimately, I think it belongs in boilerplate from an ease of use perspective, but could be convinced otherwise.

* Currently `Template` (and `Theme` too) just add a class to the body. How does this change that?

Inherently, it doesn't. It just adds some additional functionality based on the template, to allow auto-blocking, custom styling, and anything else that may be necessary to do on a per-template basis, which can help to keep styles.css and scripts.js slimmer. If we wanted to leave the `Template` metadata as it exists today alone, we could probably tie this to a different metadata field altogether (page type?), but I think that risks adding confusion between the 3 metadata fields.


kptdobe:
@shsteimer I assume we can close this one ? Closing, feel free to reopen if needed.

PR #184, opened 2023-03-01
chore: remove cls protection from css to script

davidnuescheler:
we have some `CLS` protection in `styles.css` that are a part of `lib-franklin` and not part of the project.
to separate concerns a little better, i am proposing to move those directly into `lib-franklin` i think there is a possibly for a `FOUC` but i would say that possibility has existed before, and some PSI testing may shed some light...

https://cls-to-script--helix-project-boilerplate--adobe.hlx.page/

vs.

https://main--helix-project-boilerplate--adobe.hlx.page/


scripts/lib-franklin.js
```diff
@@ -322,14 +324,15 @@ export function updateSectionsStatus(main) {
   const sections = [...main.querySelectorAll(':scope > div.section')];
   for (let i = 0; i < sections.length; i += 1) {
     const section = sections[i];
-    const status = section.getAttribute('data-section-status');
+    const status = section.dataset.sectionStatus;
     if (status !== 'loaded') {
       const loadingBlock = section.querySelector('.block[data-block-status="initialized"], .block[data-block-status="loading"]');
       if (loadingBlock) {
-        section.setAttribute('data-section-status', 'loading');
+        section.dataset.sectionStatus = 'loading';
```
bstopp, line 331: Can a block skip straight to loading status, bypassing the `initialized` section status, which hides the block until it's loaded?

Meaning: should this section _also_ set the display to none, or is that unnecessary?

davidnuescheler: yeah, blocks are not hidden... this is mostly due to the fact that sections really cannot be displayed before all blocks are loaded, due to the `<block>-container` on sections, which i really don't like anymore, but removing that is a different story.

for now, the unhiding works on a section by section basis, and i don't think we should change that behavior in this PR.

davidnuescheler: scratch my previous comment... 
i didn't read things correctly...

yes, no section can go to `loading` without being `initialized` first

bstopp: Gotcha.

bstopp (approved):

fkakatie (approved):
I think this is so much clearer. 🎉

PR #192, opened 2023-03-16
fix: svg polygons not showing up; using DOMParser for svg

### LINKED ISSUE #191: decorateIcons() accidentally hides polygons from SVG elements
### Expected Behaviour
SVG images, including vector logos, displaying correctly all the elements (e.g. rects)

### Actual Behaviour
some SVG images are shown with missing rects or polygons

### Root cause (identified)
The following lines are stripping width and height attributes from all the vector elements in a svg element:
https://github.com/adobe/helix-project-boilerplate/blob/main/scripts/lib-franklin.js#LL161-L162

In addition, regexp are used to strip or replace attributes and their values: a safer approach (proposed in the linked PR) would be to use an instance of `DOMParser`.

### Proposed solution
- The PR for this fix removes width and height attributes only from the root <svg> element.
- Usage of DOMParser where possible (as mentioned)

dicagno:
fix: svg polygons not showing up; using DOMParser instead of regex for svg manipulation, for improved safety

Fix #191 

Test URL: https://fix-decorated-svg--helix-project-boilerplate--dicagno.hlx.page/

### DIFF
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -163,7 +163,7 @@ export async function decorateIcons(element) {
           for (let i = 0; i < attributes.length; i += 1) {
             // attributes not to be transferred to the new <symbol> elem
             const skippedAttrs = ['xmlns'];
-            if (skippedAttrs.includes(attributes[i].name)) {
+            if (!skippedAttrs.includes(attributes[i].name)) {
               svgSymbol.setAttribute(attributes[i].name, attributes[i].value);
             }
           }
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -149,7 +149,7 @@ export async function decorateIcons(element) {
     if (!ICONS_CACHE[iconName]) {
       ICONS_CACHE[iconName] = true;
       try {
-        const response = await fetch(`${window.hlx.codeBasePath}${window.hlx.codeBasePath}/icons/${iconName}.svg`);
+        const response = await fetch(`${window.hlx.codeBasePath}/icons/${iconName}.svg`);
         const svgSource = await response.text();
         if (svgSource.match(/(<style | class=)/)) {
           ICONS_CACHE[iconName] = { styled: true, html: svgSource };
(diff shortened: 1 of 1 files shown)


scripts/lib-franklin.js
```diff
@@ -141,32 +141,42 @@ export async function decorateIcons(element) {
...
-        const response = await fetch(`${window.hlx.codeBasePath}/icons/${iconName}.svg`);
-        if (!response.ok) {
-          ICONS_CACHE[iconName] = false;
-          return;
-        }
-        // Styled icons don't play nice with the sprite approach because of shadow dom isolation
-        const svg = await response.text();
-        if (svg.match(/(<style | class=)/)) {
-          ICONS_CACHE[iconName] = { styled: true, html: svg };
+        const response = await fetch(`${window.hlx.codeBasePath}${window.hlx.codeBasePath}/icons/${iconName}.svg`);
```
herzog31, line 152:
```suggestion
        const response = await fetch(`${window.hlx.codeBasePath}/icons/${iconName}.svg`);
```

herzog31 (approved):

scripts/lib-franklin.js
```diff
@@ -141,32 +141,42 @@ export async function decorateIcons(element) {
...
+        const svgSource = await response.text();
+        if (svgSource.match(/(<style | class=)/)) {
+          ICONS_CACHE[iconName] = { styled: true, html: svgSource };
         } else {
+          const parser = new DOMParser();
+          const parsedSvg = parser.parseFromString(svgSource, 'image/svg+xml');
+          const newSvg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
```
ramboz, line 159: Shouldn't this rather be a `symbol` element instead of an `svg` one? The point was to create a reference template for the `<use>` tag below

ramboz (commented):
I would also try to validate this against an icon-heavy page to be sure the perf impact isn't perceptible.

https://github.com/BambooHR/bamboohr-website/pull/75 was the starting point for this initial work, so you could check against their page to measure the impact.

scripts/lib-franklin.js
dicagno: Good point, I will fix it accordingly. thanks

dicagno:
Tested the latest changes on a couple of different use cases, some svg elements are not displayed. Working on fixing this.

herzog31 (approved):

PR #202, opened 2023-03-31
Preload the Hero Image

### LINKED ISSUE #201: Preload Hero Image
Many PSI reports show a warning about preloading the FCP image to improve the user experience.

PageSpeed_Insights

I figure this should come with the boilerplate.

bstopp:
Fix #201 

Test URLs:
- After: https://feat-preload-hero--helix-project-boilerplate--bstopp.hlx.page/

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -16,6 +16,23 @@ import {
 const LCP_BLOCKS = []; // add your LCP blocks to the list
 window.hlx.RUM_GENERATION = 'project-1'; // add your RUM generation information here
 
+function preloadHeroImage(picture) {
+  const src = [...picture.querySelectorAll('source')]
+    .filter((source) => source.getAttribute('type') === 'image/webp')
+    .find((source) => {
+      const media = source.getAttribute('media');
+      return !media || window.matchMedia(media).matches;
+    });
+
+  const link = document.createElement('link');
+  link.setAttribute('rel', 'preload');
+  link.setAttribute('fetchpriority', 'high');
+  link.setAttribute('as', 'image');
+  link.setAttribute('href', src.getAttribute('srcset'));
+  link.setAttribute('type', src.getAttribute('type'));
+  document.head.append(link);
+}
+
 /**
  * Builds hero block and prepends to main in a new section.
  * @param {Element} main The container element
@@ -25,6 +42,7 @@ function buildHeroBlock(main) {
   const picture = main.querySelector('picture');
   // eslint-disable-next-line no-bitwise
   if (h1 && picture && (h1.compareDocumentPosition(picture) & Node.DOCUMENT_POSITION_PRECEDING)) {
+    preloadHeroImage(picture);
     const section = document.createElement('div');
     section.append(buildBlock('hero', { elems: [picture, h1] }));
     main.prepend(section);


bstopp:
If anyone knows a better way to force the browser to process the picture element, _without it being visible_, i'm all for it. But i couldn't figure out how to make that event take place.

ramboz:
You could create a specific preload Link header for the pages via the global headers spreadsheet. But ideally, we should even have the pipeline do that for us based on the image metadata for the page.

@davidnuescheler @tripodsan @kptdobe  what do you guys think?

bstopp:
That assumes the image metadata is the same as what needs to be preloaded for hero. But that is typically the case. 

Also the preload image need for mobile is diff than desktop, and diff from the metadata.

ramboz:
The `Link` headers seem to support media queries, if I understand the discussion here correctly https://github.com/w3c/preload/issues/18 and the spec https://html.spec.whatwg.org/multipage/links.html#link-type-preload. But it does require that those links be manually set in the spreadsheet.

If this does not work, then we'd be indeed stuck with only supporting a fixed URL. The `<link>` meta prop does support the media prop for sure, though, but I don't think the pipeline would let you set this from the document metadata directly. And if you have to do it in JS, we are back to your solution.

Now the complaint is mostly on the mobile PSI score, so in theory, we could have it hardcoded in the metadata, or the document, manually for now. You would likely get a warning on desktop that you asked to preload something that wasn't used in the first 3s though… but since bandwidth is usually better on desktop, this is less of an issue. It's a trade-off in the end if you really need to "over-optimize" for mobile.

But before even trying further, I would try to set it manually once to see the PSI score difference, maybe using @davidnuescheler 's https://thinktanked.org/tools/deep-psi/deep-psi.html so you get a good average. I'm not sure it would really change the score much anyways.

bstopp:
You lost me on some of that - but my understanding is that this particular item on the PSI report has no impact on the actual score. It is more of an information 'You should do this to make a better experience for your users'.

ramboz:
I think the `Link` headers, `<link>` tags in the `<head>` and `fetchpriority` on the image are the only options we have to control the loading sequence. Each approach will have pros and cons, and it will likely be a project-specific implementation detail to squeeze the last few ms out of the LCP.

I also remember from discussions with @davidnuescheler that those kind of optimizations tend to work well for desktop, but ended impacting mobile negatively on several test projects as the bandwidth is more limited and the overhead of handling several requests in parallel was larger than the gain from preloading earlier. At least on "lab" measurements we did via PSI. It would need to be confirmed to "field" data from CrUX to be sure.

rofe:
Unfortunately PSI recommendations aren't always correct or useful...

@bstopp Have you seen this becoming an actual issue in a real life scenario? In the boilerplate project, I can't see it moving the needle at all, on the contrary:
### main
image

### fork
image

bstopp:
> No, I haven't seen it have any meaningful impact on the score itself. But that the report says "You should fix this", to me in something to consider.

I should clarify : i haven't seen it show any impact on the Boilerplate score. I'd have to remove it from my projects to test those contexts.

davidnuescheler:
> This happens when the LCP is finally 'visible' to the browser and in the DOM ("loaded").
no, that's not how it works. the body actually only get unhidden when the LCP candidate is loaded. so it is being loaded when it is `display: none` and the loading starts with the LCP being set to `loading="eager"` (which happens super early in the decoration phase).
we originally had the `loading="eager"` on the server, but that means that you have to guess your LCP on the server, which can be wrong, especially if you load the actual `LCP` element from a JSON (eg. an article feed), and moving it to the client made everything much more flexible and simple without any impact on performance.

davidnuescheler:
@bstopp i am going to close this one, please feel free to reopen if you still think that this is an issue.

PR #204, opened 2023-04-06
fix: Unhandled Promise Rejection in lib-franklin fetchPlaceholders 

### LINKED ISSUE #203: Unhandled Promise Rejection in lib-franklin fetchPlaceholders
There is a potential for an Unhandled Promise Rejection in the fetchPlaceholders of lib-franklin. The promise is wrapping the fetch call in a try/catch but as there is no await, it should instead add a catch at the end of the chain. Otherwise, a rejected fetch results in an unhandled promise.

Furthermore, the response status of the fetch is not checked which will make it potentially fail on the .json call in case of a request with a response that has a status other than ok. 

Additionally, the resulting json is assumed to have entries with Key and Text in the data section - if that is not the case, the resulting object has undefined keys.

### Expected Behaviour
A rejected fetch should not result in an unhandled promise and a fetch with a status that is not ok should result in a promise rejection. Additionally, the final placeholders object should not contain undefined keys.

### Actual Behaviour
A rejected fetch causes an Unhandled Promise Rejection. A fetch with a status that is not ok errors on the .json call. A result with entries without Key in the data section adds undefined.

### Reproduce Scenario (including but not limited to)
A call to fetchPlaceholders where there is no placeholders sheet. A status code returned other than 200. A mistake in the sheet where no Key is given.

#### Steps to Reproduce
For the main cases (the Unhandled Promise Rejection) it is enough to do a fetchPlaceholders where there is no placeholders sheet set-up.

#### Sample Code that illustrates the problem
```
import { fetchPlaceholders } from '../../scripts/lib-franklin.js';
....
const placeholders = await fetchPlaceholders(`/en`);
```
#### Logs taken while reproducing problem
```
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (placeholders.json, line 0)
[Error] Unhandled Promise Rejection: SyntaxError: The string did not match the expected pattern.
	promiseEmptyOnRejected
	promiseReactionJob
```

karlpauls:
Fix #203

Test URLs:
- Before: https://main--helix-project-boilerplate--adobe.hlx.live
- After: https://203--helix-project-boilerplate--karlpauls.hlx.live

### DIFF
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -196,22 +196,26 @@ export async function fetchPlaceholders(prefix = 'default') {
   const loaded = window.placeholders[`${prefix}-loaded`];
   if (!loaded) {
     window.placeholders[`${prefix}-loaded`] = new Promise((resolve, reject) => {
-      try {
-        fetch(`${prefix === 'default' ? '' : prefix}/placeholders.json`)
-          .then((resp) => resp.json())
-          .then((json) => {
-            const placeholders = {};
-            json.data.forEach((placeholder) => {
+      fetch(`${prefix === 'default' ? '' : prefix}/placeholders.json`)
+        .then((resp) => {
+          if (resp.ok) {
+            return resp.json();
+          }
+          throw new Error(`${resp.status}: ${resp.statusText}`);
+        }).then((json) => {
+          const placeholders = {};
+          json.data
+            .filter((placeholder) => placeholder.Key && placeholder.Text)
+            .forEach((placeholder) => {
               placeholders[toCamelCase(placeholder.Key)] = placeholder.Text;
             });
-            window.placeholders[prefix] = placeholders;
-            resolve();
-          });
-      } catch (error) {
-        // error loading placeholders
-        window.placeholders[prefix] = {};
-        reject();
-      }
+          window.placeholders[prefix] = placeholders;
+          resolve();
+        }).catch((error) => {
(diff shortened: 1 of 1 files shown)


trieloff (approved):

rofe (approved):

scripts/lib-franklin.js
```diff
@@ -196,22 +196,26 @@ export async function fetchPlaceholders(prefix = 'default') {
...
+      fetch(`${prefix === 'default' ? '' : prefix}/placeholders.json`)
+        .then((resp) => {
+          if (resp.ok) {
+            return resp.json();
+          }
+          throw new Error(`${resp.status}: ${resp.statusText}`);
+        }).then((json) => {
+          const placeholders = {};
+          json.data
+            .filter((placeholder) => placeholder.Key && placeholder.Text)
```
rofe, line 208: Filtering for existing `Key` is certainly useful, but making `Text` "truthiness" mandatory might be breaking backward compatibility: I think placeholders may also be empty strings under certain circumstances.

karlpauls: Thanks, that makes sense - I think it is not important to filter based on an existing `Text`. The filtering on existing `Key` is the important bit. @rofe, do you want me to remove the `&& placeholder.Text` from the filter?

rofe: yes, please 🙏

PR #206, opened 2023-04-11
feat(rum): allow masking URLs sent to RUM service to preserve visitor privacy

trieloff:
in some cases, the URL may contain PII as part of the query string or the path. To
avoid collecting PII through RUM, you can now set the `window.hlx.RUM_MASK_URL` property to one of
- `full`: collect the full URL, this is the previous behavior and new default
- `path`: collects the URL, but ignores all URL parameters
- `origin`: collects only the hostname, not the path of the URL

BREAKING CHANGE: the `generation` parameter is no longer collected as part of RUM data

Fix #<gh-issue-id>

Test URLs:
- Before: https://main--<repo>--<owner>.hlx.page/
- After: https://<branch>--<repo>--<owner>.hlx.page/

### DIFF
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -41,14 +41,19 @@ export function sampleRUM(checkpoint, data = {}) {
       const id = `${hashCode(window.location.href)}-${new Date().getTime()}-${Math.random().toString(16).substr(2, 14)}`;
       const random = Math.random();
       const isSelected = (random * weight < 1);
-      // eslint-disable-next-line object-curly-newline
-      window.hlx.rum = { weight, id, random, isSelected, sampleRUM };
+      const urlSanitizers = {
+        full: () => window.location.href,
+        origin: () => window.location.origin,
+        path: () => window.location.href.replace(/\?.*$/, ''),
+      };
+      // eslint-disable-next-line object-curly-newline, max-len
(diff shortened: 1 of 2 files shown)


rofe (approved):
Why are we dropping `generation` at the same time as adding URL sanitization?

scripts/lib-franklin.js
```diff
@@ -41,14 +41,19 @@ export function sampleRUM(checkpoint, data = {}) {
       const id = `${hashCode(window.location.href)}-${new Date().getTime()}-${Math.random().toString(16).substr(2, 14)}`;
       const random = Math.random();
       const isSelected = (random * weight < 1);
-      // eslint-disable-next-line object-curly-newline
-      window.hlx.rum = { weight, id, random, isSelected, sampleRUM };
+      const urlSanitizers = {
+        full: () => window.location.href,
+        origin: () => window.location.origin,
+        path: () => window.location.href.replace(/\?.*$/, ''),
```
ramboz, line 47: Any reason not to us `window.location.pathname`?

trieloff:
```suggestion
        path: () => window.location.pathname,
```

trieloff:
I can split the drop of `generation` into a separate PR, but given its lack of use, it is pretty unimportant.

PR #210, opened 2023-04-24
feat: Dom Helpers for more concise block code

andreituicu:
## Context
This PR adds a small JS file to help with DOM elements generation.
We've been using this in our project and it speeds up a lot the development, code reviews and testing, so I thought it might be useful for new projects as well.

https://github.com/hlxsites/moleculardevices/pull/161

## Why?
The following Dom Helper functions allow for:
* a more concise code
* composable structure
* dom like syntax structure to easily visualise the resulting HTML when looking at the code
* faster development and code review of block code
* while maintaining  100 LHS for performance
* 100% vanilla JS / no compilation required
* with minimal overhead - a few ifs and function calls

## Main Usecases
* DOM rendering of dynamic data from spreadsheets and indexes
* Adding new DOM elements during block decoration to what Franklin provides, based on the word document (e.g. adding buttons for a carousel, where the elements are coming from the word document)

## Examples
It allows rewrite from:
```javascript
function renderItem(item) {
  const newsItem = document.createElement('div');
  newsItem.classList.add('blog-carousel-item');
  const newsItemLink = document.createElement('a');
  newsItemLink.href = item.path;
  newsItem.append(newsItemLink);
  const newsThumb = document.createElement('div');
  newsThumb.classList.add('blog-carousel-thumb');
  newsThumb.append(createOptimizedPicture(item.image, item.title, 'lazy', [{ width: '800' }]));
  newsItemLink.appendChild(newsThumb);
  const newsCaption = document.createElement('div');
  newsCaption.classList.add('blog-carousel-caption');
  newsCaption.innerHTML = `
    <h3>${item.title}</h3>
    <p class="blog-description">${item.description}</p>
    <p class="button-container">
      <a href=${item.path} aria-label="Read More" class="button primary">Read More</a>
    </p>
  `;
  newsItemLink.appendChild(newsCaption);
  return newsItem;
}
```

To:
```javascript
function renderItem(item) {
  return (
    div({ class: 'blog-carousel-item' },
      a({ href: item.path },
        div({ class: 'blog-carousel-thumb' },
          createOptimizedPicture(item.image, item.title, 'lazy', [{ width: '800' }]),
        ),
        div({ class: 'blog-carousel-caption' },
          h3(item.title),
          p({ class: 'blog-description' }, item.description),
          p({ class: 'button-container' },
            a({ href: item.path, 'aria-label': 'Read More', class: 'button primary' }, 'Read More'),
          ),
        ),
      ),
    )
  );
}
```

More examples from our project code:
* https://github.com/hlxsites/moleculardevices/blob/main/templates/blog/blog.js#L57-L74
* https://github.com/hlxsites/moleculardevices/blob/main/blocks/related-assay-data/related-assay-data.js#L33-L41
* https://github.com/hlxsites/moleculardevices/blob/main/blocks/social-share/social-share.js#L94-L106
* https://github.com/hlxsites/moleculardevices/blob/main/blocks/citations/citations.js#L83-L110
* https://github.com/hlxsites/moleculardevices/blob/main/blocks/social-share/social-share.js#L26-L34

## Other projects

The reason why I thought it would be a good addition to the boiler plate is that I found similar/partial implementations, some repeating between projects, some independently written.

1. Adobe.com: https://blog.adobe.com/blocks/block-helpers.js
2. Volvo Trucks: https://github.com/hlxsites/vg-volvotrucks-us/blob/main/scripts/scripts.js#L40-L53
3. Caesars: https://github.com/hlxsites/caesars/blob/main/caesars-palace/scripts/scripts.js#L244-L263
4. Airlessco: https://github.com/hlxsites/airlessco/blob/main/scripts/scripts.js#L119-L134
5. WKND: https://github.com/hlxsites/wknd/blob/main/scripts/scripts.js#L50-L69 
6. Mammotome: https://github.com/hlxsites/mammotome/pull/67/files#diff-a5ee87ff9063f921ba01933977e6164e5729551986030f0ed7eb747bcffb740fR228 (Thank you, @karlpauls!)

 
## Credits

This is of course inspired by the JSX (React) syntax, articles and code pieces explaining its internals.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,5 +17,6 @@ module.exports = {
     'import/extensions': ['error', {
       js: 'always',
     }],
+    'function-paren-newline': 'off',
   },
 };
--- a/scripts/dom-helpers.js
+++ b/scripts/dom-helpers.js
@@ -1,0 +1,76 @@
+/* eslint-disable no-param-reassign */
+
+/**
+ * Example Usage:
+ *
+ * domEl('main',
+ *  div({ class: 'card' },
+ *  a({ href: item.path },
+ *    div({ class: 'card-thumb' },
+ *     createOptimizedPicture(item.image, item.title, 'lazy', [{ width: '800' }]),
+ *    ),
+ *   div({ class: 'card-caption' },
+ *      h3(item.title),
+ *      p({ class: 'card-description' }, item.description),
+ *      p({ class: 'button-container' },
+ *       a({ href: item.path, 'aria-label': 'Read More', class: 'button primary' }, 'Read More'),
+ *     ),
+ *   ),
+ *  ),
+ * )
+ */
+
+/**
+ * Helper for more concisely generating DOM Elements with attributes and children
+ * @param {string} tag HTML tag of the desired element
+ * @param  {[Object?, ...Element]} items: First item can optionally be an object of attributes,
+ *  everything else is a child element
+ * @returns {Element} The constructred DOM Element
+ */
+export function domEl(tag, ...items) {
+  const element = document.createElement(tag);
+
+  if (!items || items.length === 0) return element;
+
+  if (!(items[0] instanceof Element || items[0] instanceof HTMLElement) && typeof items[0] === 'object') {
+    const [attributes, ...rest] = items;
+    items = rest;
+
+    Object.entries(attributes).forEach(([key, value]) => {
+      if (!key.startsWith('on')) {
+        element.setAttribute(key, Array.isArray(value) ? value.join(' ') : value);
+      } else {
+        element.addEventListener(key.substring(2).toLowerCase(), value);
+      }
+    });
+  }
+
+  items.forEach((item) => {
+    item = item instanceof Element || item instanceof HTMLElement
+      ? item
+      : document.createTextNode(item);
+    element.appendChild(item);
+  });
+
+  return element;
+}
+
+/*
+  More short hand functions can be added for very common DOM elements below.
+  domEl function from above can be used for one off DOM element occurrences.
+*/
+export function div(...items) { return domEl('div', ...items); }
+export function p(...items) { return domEl('p', ...items); }
+export function a(...items) { return domEl('a', ...items); }
+export function h1(...items) { return domEl('h1', ...items); }
+export function h2(...items) { return domEl('h2', ...items); }
+export function h3(...items) { return domEl('h3', ...items); }
+export function h4(...items) { return domEl('h4', ...items); }
+export function h5(...items) { return domEl('h5', ...items); }
+export function h6(...items) { return domEl('h6', ...items); }
+export function ul(...items) { return domEl('ul', ...items); }
+export function li(...items) { return domEl('li', ...items); }
+export function i(...items) { return domEl('i', ...items); }
+export function img(...items) { return domEl('img', ...items); }
+export function span(...items) { return domEl('span', ...items); }
+export function button(...items) { return domEl('button', ...items); }
--- a/test/scripts/dom-helpers.test.js
+++ b/test/scripts/dom-helpers.test.js
@@ -1,0 +1,88 @@
+/* global describe it */
+import { expect } from '@esm-bundle/chai';
+import {
+  domEl, div, a, p, h3, button,
+} from '../../scripts/dom-helpers.js';
+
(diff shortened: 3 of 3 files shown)


karlpauls:
👍 - @solaris007 and I are working on something similar (arguably not as sophisticated as this proposal) in [0]. I think it would make some sense to have a standard way to create nested Dom structures in a readable way.

[0] https://github.com/hlxsites/mammotome/pull/67/files#diff-a5ee87ff9063f921ba01933977e6164e5729551986030f0ed7eb747bcffb740fR228

dylandepass:
@andreituicu I'm curious what limitations in template literals made you come up with this approach?

One benefit I can see is it's easier to attach event listeners and maybe less need for innerHTML, but besides that I'm not sure I see what else you get. 

I would 100% agree that it's much easier to read and maintain than imperative DOM code. But on the flip side I think the same can be done with string template literals (and still have it be concise. composable and easy to visualize).

```js
function renderItem(item) {
  return /* html */`
    <div class='blog-carousel-item'>
      <a href=${item.path}>
        <div class='blog-carousel-thumb'>
          ${createOptimizedPicture(item.image, item.title, 'lazy', [{ width: '800' }])}
        </div>
        <div class='blog-carousel-caption'>
          <h3>${item.title}</h3>
          <p class='blog-description'>${item.description}</p>
          <p class='button-container'>
            <a href=${item.path} aria-label='Read More' class='button primary'>Read More</a>
          </p>
        </div>
      </a>
    </div>
  `;
}
```

trieloff:
To me, this looks like a templating language and templating languages should be a user choice, not a framework default.

The most vanilla thing that I could imagine is based on the HTML template element https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template – but it's not very widely adopted yet.

andreituicu:
@dylandepass 

> Correct, my bad, I through the example together quickly without thinking too much about it. I solved this in the past by using outerHTML as you mentioned.

Sure, I should've maybe been explicit: I only pointed it to show that, although possible with string literals, the developer experience when writing it could be a little bit more seamless from my point of view. 

> One note, you can fix the syntax highlighting and code completion issue in vscode with [this](https://marketplace.visualstudio.com/items?itemName=Tobermory.es6-string-html) extension.

Noted! Was not aware! Thank you!

> I just wanted to better understand the pain points here and see if there was something else I was missing.

Happy to discuss and thank you very much for the feedback!

mhaack:
While I see the point on not putting this into boilerplate, as @trieloff pointed out it should be a user's choice.

There seems to be a need for this. I'm aware two other projects built helpers similar like this one. We might add this to the block party instead of boilerplate.

andreituicu:
@mhaack Thank you for the feedback!

> We might add this to the block party instead of boilerplate. 

you are referring to https://github.com/adobe/helix-block-collection , correct?
I can open a PR there if that's a more appropriate place for this and contribute a doc entry to https://www.hlx.live/developer/block-collection#block-collection to make it easier discoverable.

@trieloff does that sounds like a good approach to you?

>  I'm aware two other projects built helpers similar like this one.

just checking: are they captured in the **Other Projects** section in this PR? if not could you please share them so I can add those as well?

davidnuescheler:
i agree that the best place would be "block party" mentioned by @mhaack to get started.

the "block party" is a new concept we came up with for community contributed blocks and snippets. i don't think the code should be in any particular repos but it lives where it is... i don't know the latest on process for including things in the block party. maybe it makes sense to track those as issues in https://github.com/adobe/helix-website. 
(cc: @amol-anand)

trieloff:
https://github.com/hlxsites/block-party seems like the best place for this.

davidnuescheler:
should we move this issue to `block-party` and close this `PR`?

andreituicu:
@davidnuescheler @trieloff Thank you! Created https://github.com/hlxsites/block-party/pull/1 . 

Do you mind if I also keep this PR open as well for discoverability reasons? At the moment, not that many people know about the Block Party repository, while I think many check the boilerplate pull requests for possible improvemennts.

trieloff:
To solve the discoverability issue, you could submit a PR that updates the README with links to block collection and block party.

andreituicu:
@trieloff just to clarify I understood correctly: You mean a PR to the README of this repository (the boilerplate)? Sure, will do!

PR #211, opened 2023-04-25
chore: remove test setup

rofe:
Consensus among project leads is that the current test setup is creating more work than it is generating value for projects.

https://remove-tests--helix-project-boilerplate--adobe.hlx.page/


davidnuescheler (approved):

rofe:
@dylandepass maybe, but we aren't currently testing anything useful in the current test setup. Basically just `header` and `footer` tests which break immediately after the first commit to those files in a project... the perceived value of unit-testing blocks among project leads is zero, even if only in boilerplate.

On the other hand, we aren't testing anything in `lib-franklin.js` which should ideally have 100% test coverage. The solution is probably to move `lib-franklin.js` into its own repo and add it as an NPM dependency?

dylandepass:
> On the other hand, we aren't testing anything in lib-franklin.js which should ideally have 100% test coverage. The solution is probably to move lib-franklin.js into its own repo and add it as an NPM dependency?

Fair point, I just took a closer look at the tests and you are correct. Not much of significance is in there and nothing covering `lib-franklin.js`.

I'm totally in agreement about moving `lib-franklin.js` to a package and having the test coverage there.

rofe:
> Not much of significance is in there

@dylandepass the "real" home of the blocks in this repo is `helix-block-collection` anyway, where we still have the same test setup. We could look into better ways to test blocks there.

> Do the test commands defined in package.json also need to go as a part of this PR?
@sdmcraft good catch! I had removed them locally, but forgot to save 😬

PR #212, opened 2023-04-26
use modern API in addFavIcon

synox:
Based on https://javascript.info/modifying-document#summary these are also “old school” methods.

Simplifies the code, easier to understand.

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -86,6 +86,7 @@ export function addFavIcon(href) {
     existingLink.replaceWith(link);
   } else {
     document.head.appendChild(link);
+  }
 }
 
 /**


scripts/scripts.js
```diff
@@ -83,10 +83,9 @@ export function addFavIcon(href) {
   link.href = href;
   const existingLink = document.querySelector('head link[rel="icon"]');
   if (existingLink) {
-    existingLink.parentElement.replaceChild(link, existingLink);
+    existingLink.replaceWith(link);
   } else {
-    document.getElementsByTagName('head')[0].appendChild(link);
-  }
```
rofe, line 89: I think this bracket should have stayed

rofe (changes requested):

scripts/scripts.js
synox: oops! fixed

synox:
@rofe done

rofe:
@synox great, now all that's left to do is adding the test URL to the description for the last check to turn green.

rofe:
Changes included in #214.

PR #213, opened 2023-05-02
Update lib-franklin.js to fix issue with snake case icons

travrob1:
HI

This pr fixes and issue with snake cased icons..ie 
chevron-left.svg

I don't have a specific helix boilderplate ticket.  We ran into this issue in our project during this issue.

https://github.com/hlxsites/petplace/issues/1

https://patch-1--helix-project-boilerplate--travrob1.hlx.page/


trieloff (changes requested):
I think the prior version was more concise.

travrob1:
hi @trieloff , The issue was if there were multiple `-` in a name.

`chevron-left.svg` for example.  If we are justing using the previous logic, then the name we end up with just `chevron`.

This logic is used a couple [of lines above ](https://github.com/adobe/helix-project-boilerplate/pull/213/files#diff-a5ee87ff9063f921ba01933977e6164e5729551986030f0ed7eb747bcffb740fR149)... but was not applied here at this location

dylandepass:
I also recently ran into this issue. In the end I just removed the dash from the icon name (which is less than ideal). So I support this change 🙏🏽

trieloff (commented):
As the function is `export`ed, a test case should be easy to provide and would be a good practice in illustrating the use case and why the fix is effective.

dylandepass:
@trieloff The test setup was removed with https://github.com/adobe/helix-project-boilerplate/pull/211

trieloff (approved):

PR #214, opened 2023-05-08
chore: more realistic, working ootb favicon

davidnuescheler:
https://favicon-update--helix-project-boilerplate--adobe.hlx.live/
vs.


scripts/scripts.js
```diff
@@ -105,7 +105,7 @@ async function loadLazy(doc) {
   loadFooter(doc.querySelector('footer'));
 
   loadCSS(`${window.hlx.codeBasePath}/styles/lazy-styles.css`);
-  addFavIcon(`${window.hlx.codeBasePath}/styles/favicon.svg`);
+  addFavIcon(`${window.hlx.codeBasePath}/styles/favicon.png`);
```
rofe, line 108: Should we use the opportunity to move the favicon to the `icons` folder?

rofe (approved):

scripts/scripts.js
davidnuescheler: i think that the fact that this is not a `.svg` and that you couldn't use it in the `:icon:` notation kind of wants me to keep it away from there.

PR #215, opened 2023-05-08
font example

davidnuescheler:
the vast majority of projects load custom web fonts. font loading while preserving a great user experience can be complex and often is dependent on the specific fonts that are being used. 
this PR provides an example that should work for most projects, and combines and initial lazy load on mobile, to avoid impact to `LCP` and loads fonts eager on subsequent pages on mobile. on desktop there is usually more bandwidth so loading fonts early can probably happen without material `LCP` impact (depending on the font).

there are a number of open questions here:
(1) where should we put licensing terms for roboto
(2) the font-loading sequence should probably be related to bandwidth not screen size
(3) putting roboto into the repo, means that most project will have to remove those files and replace them with other self hosted font files

https://font-example--helix-project-boilerplate--adobe.hlx.live/
vs.


davidnuescheler:
at least for boilerplate it looks like there is no discernible performance impact. 

Screenshot 2023-05-07 at 6 28 30 PM

Screenshot 2023-05-07 at 6 28 59 PM

scripts/scripts.js
```diff
@@ -105,6 +109,11 @@ async function loadLazy(doc) {
   loadFooter(doc.querySelector('footer'));
 
   loadCSS(`${window.hlx.codeBasePath}/styles/lazy-styles.css`);
+
+  loadCSS(`${window.hlx.codeBasePath}/styles/fonts.css`, () => {
+    if (!window.location.hostname.includes('localhost')) sessionStorage.setItem('fonts-loaded', 'true');
```
kptdobe, line 114: Why `sessionStorage` and not `localStorage` ? If I frequently come back on the site on mobile (and close my browser app), I'll have to wait before the font is loaded while the font is cached already.

scripts/scripts.js
```diff
@@ -70,6 +70,10 @@ async function loadEager(doc) {
     document.body.classList.add('appear');
     await waitForLCP(LCP_BLOCKS);
   }
+  if (window.innerWidth >= 900) loadCSS(`${window.hlx.codeBasePath}/styles/fonts.css`);
+  if (sessionStorage.getItem('fonts-loaded')) {
+    loadCSS(`${window.hlx.codeBasePath}/styles/fonts.css`);
+  }
```
kptdobe, line 76: This is calling `loadCSS` twice on desktop the 2nd time... browser is probably smart but could be optimised as:

```suggestion
  if (window.innerWidth >= 900 || sessionStorage.getItem('fonts-loaded')) {
    // load fonts.css immediately on desktop or if font has already been loaded
    loadCSS(`${window.hlx.codeBasePath}/styles/fonts.css`);
  }
```

scripts/scripts.js
kptdobe: Loading fonts.css a 2nd time on desktop... again, this probably works but not necessarily nice. We could use a global flag to track if we have loaded fonts.css or not. And also set the sessionStorage property in the eager phase too.

kptdobe (changes requested):

scripts/scripts.js
davidnuescheler: `localStorage` is permanent, by default we cache for 2h for items coming from the code bus which seems much more of a `session` scope...

davidnuescheler: i think `loadCSS()` would not add the `fonts.css` if it is already present, and if it did, the browser wouldn't load it twice either, so i think this is pretty safe.

scripts/scripts.js
davidnuescheler: i am not sure the extra code is needed given that `loadCSS()` already checks

scripts/scripts.js
```diff
@@ -70,6 +70,10 @@ async function loadEager(doc) {
     document.body.classList.add('appear');
     await waitForLCP(LCP_BLOCKS);
```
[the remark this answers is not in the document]
davidnuescheler: this is not about blocking but instead about bandwidth constraint... so if you have multiple requests go on in parallel, they will share the bandwidth, in which case the loading of the `LCP` would take 2x as long.

scripts/scripts.js
arumsey: `sessionStorage` may throw an exception if quota has been exceeded or user disabled web storage. You may want to wrap in a try catch.

davidnuescheler: good point. i have definitely never done that in the past, but i am not opposed to it.

fkakatie (approved):

scripts/scripts.js
```diff
@@ -70,6 +82,15 @@ async function loadEager(doc) {
...
   }
+
+  try {
+    /* if desktop (proxy for fast connection) or fonts already loaded, load fonts.css */
```
icaraps, line 87: > fast connection

It's not widely supported yet but you could use `navigator.connection.downlink` to get the bandwidth estimate in megabits per second to know if the user has a fast connection or not https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink

PR #216, opened 2023-05-08
chore: refactor loadCSS, add loadScript

davidnuescheler:
this is an attempt to modernize `loadCSS()` and add `loadScript()`
for loadCSS this is a breaking change, input is very welcome...

i lifted a lot of the `loadScript()` code from @synox 

https://load-css-script--helix-project-boilerplate--adobe.hlx.live/
vs.

### DIFF
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -82,19 +82,41 @@ export function sampleRUM(checkpoint, data = {}) {
  * Loads a CSS file.
  * @param {string} href The path to the CSS file
  */
-export function loadCSS(href, callback) {
-  if (!document.querySelector(`head > link[href="${href}"]`)) {
-    const link = document.createElement('link');
-    link.setAttribute('rel', 'stylesheet');
-    link.setAttribute('href', href);
-    if (typeof callback === 'function') {
-      link.onload = (e) => callback(e.type);
-      link.onerror = (e) => callback(e.type);
+export async function loadCSS(href) {
+  return new Promise((resolve, reject) => {
+    if (!document.querySelector(`head > link[href="${href}"]`)) {
+      const link = document.createElement('link');
+      link.rel = 'stylesheet';
+      link.href = href;
+      link.onload = resolve;
+      link.onerror = reject;
+      document.head.append(link);
+    } else {
+      resolve();
     }
-    document.head.appendChild(link);
-  } else if (typeof callback === 'function') {
-    callback('noop');
-  }
+  });
+}
+
+/**
+ * Loads a non module JS file.
+ * @param {string} href The path to the CSS file
+ * @param {Object} attrs additional optional attributes
+ */
+
+export async function loadScript(url, attrs) {
+  return new Promise((resolve, reject) => {
+    const script = document.createElement('script');
+    script.src = url;
+    if (attrs) {
+    // eslint-disable-next-line no-restricted-syntax, guard-for-in
+      for (const attr in attrs) {
+        script.setAttribute(attr, attrs[attr]);
+      }
+    }
+    script.onload = resolve;
+    script.onerror = reject;
+    document.head.append(script);
+  });
 }
 
 /**
@@ -399,9 +421,7 @@ export async function loadBlock(block) {
     block.dataset.blockStatus = 'loading';
     const { blockName } = block.dataset;
     try {
-      const cssLoaded = new Promise((resolve) => {
-        loadCSS(`${window.hlx.codeBasePath}/blocks/${blockName}/${blockName}.css`, resolve);
-      });
+      const cssLoaded = loadCSS(`${window.hlx.codeBasePath}/blocks/${blockName}/${blockName}.css`);
       const decorationComplete = new Promise((resolve) => {
         (async () => {
           try {
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -80,7 +80,7 @@ export function sampleRUM(checkpoint, data = {}) {
 
 /**
  * Loads a CSS file.
- * @param {string} href The path to the CSS file
+ * @param {string} href URL to the CSS file
  */
 export async function loadCSS(href) {
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


scripts/lib-franklin.js
```diff
@@ -82,19 +82,41 @@ export function sampleRUM(checkpoint, data = {}) {
...
  */
-export function loadCSS(href, callback) {
-  if (!document.querySelector(`head > link[href="${href}"]`)) {
-    const link = document.createElement('link');
-    link.setAttribute('rel', 'stylesheet');
-    link.setAttribute('href', href);
-    if (typeof callback === 'function') {
-      link.onload = (e) => callback(e.type);
-      link.onerror = (e) => callback(e.type);
+export async function loadCSS(href) {
```
kptdobe, line 85: Why not keeping the `callback` ?

scripts/lib-franklin.js
```diff
@@ -82,19 +82,41 @@ export function sampleRUM(checkpoint, data = {}) {
...
+  });
+}
+
+/**
+ * Loads a non module JS file.
+ * @param {string} href The path to the CSS file
+ * @param {Object} attrs additional optional attributes
+ */
+
+export async function loadScript(url, attrs) {
```
kptdobe, line 106: Why not having  a `callback` ? (and align the 2 methods)

kptdobe: Also, `loadCSS` check for existing while `loadJS` does not... maybe we should align the 2 impl

scripts/lib-franklin.js
kptdobe: oh... of course. I completely missed that part :)

rofe, line 85: There seem to be quite a few [usages](https://github.com/search?q=org%3Ahlxsites%20loadCSS(&type=code) relying on the `callback` argument, so we should probably try to keep the signature and behavior for backward compatibility, i.e. call the `callback` as well whenever we call `resolve` or `reject`.

It could be a bit confusing for new users, but we could mention in the jsdoc that `callback` is deprecated and that they should use `await` or `.then()` instead. Mozilla's `browser` API is an example where they implemented both callback and return value to keep compatibility with the more widely used `chrome` API.

rofe (changes requested):

scripts/lib-franklin.js
kptdobe: While I would initially agree... we do not have an "update" process of the boilerplate, apart from a manual process. One of the reason being... we do not want to cary over old code. New project benefit from new code while if existing projects want to upgrade, there is a little bit of manual work. I would not keep both.

stefanseifert: introducing backward-incompatible changes silently is a slippery slope, esp. as there is currently no release process, no changelog, no semver policy or similar in place that notifies users that the new version is not just a drop-in replacement. adding a note to the jsdoc only is not enough imho.

kptdobe: In the current state, the boilerplate is a template to start from, not to update from.
Introducing proper versioning has been questioned a lot and we did not identify the need so far. Do you have real life scenarii ? 
In 4 years, we had to do one major upgrade (5 or 6 projects to patch with the latest lib-franklin version). With the (simple) code base, this was a really trivial operation, even with no doc (git commit history is a changelog...).
The "cost" of maintaining backward compatibility is so much more expensive, especially for the optimal performances we are looking for (maintaining legacy code might slow down everything).

stefanseifert: i can currently only speak for myself (after migrating [adaptto-website](https://github.com/adaptto/adaptto-website)), and i like to look on updates on the central parts of the boilerplate and sync with latest changes from there in lib-franklin from time to time when i think the improvements are useful, and to not deviate too much from it to make future updates easier. maybe other projects do not care at all?
maintaining a release process and changelog for lib-franklin does not mean to always maintain backward-compatibility and sacrifice performance. you make a new major version and a note to the changelog about the breaking change, and it's obvious for everyone. that's not much effort.
so my point is: if you want to break compatibility for a good reason, go forward, but make it as transparently as possible for those who want to keep updated with the lastest improvements.

ramboz: Would it be too hard to just support the 2 APIs?
Just adding a conditional checking for `arguments.length` shouldn't be hard and would offer both a promise and callback approach, so we are backward compatible, and offer more flexibility on how devs can use the method

scripts/lib-franklin.js
ramboz: If we end up supporting callbacks for `loadCSS`, we probably want to do it as well for `loadJS` so the 2 are aligned

davidnuescheler:
> @davidnuescheler you mentioned me, not sure which code you pulled. Did you mean to cherry-pick https://github.com/adobe/helix-project-boilerplate/pull/212/files ?

i grabbed it from a project... `loadScript()` that is.

scripts/lib-franklin.js
synox: my 2¢: keep the boilerplate as modern and efficient as possible, even if this means breaking changes.

synox (approved):

rofe (approved):

kptdobe (approved):

scripts/lib-franklin.js
```diff
@@ -82,19 +82,41 @@ export function sampleRUM(checkpoint, data = {}) {
...
+  return new Promise((resolve, reject) => {
+    const script = document.createElement('script');
```
jckautzmann, line 108: Wouldn't it make sense to check it the script tag already exist in the head (similar to above code for CSS link)?

davidnuescheler: yeah, i think that would be great. added.

scripts/lib-franklin.js
```diff
@@ -80,21 +80,47 @@ export function sampleRUM(checkpoint, data = {}) {
...
-  }
+  });
+}
+
+/**
+ * Loads a non module JS file.
+ * @param {string} src URL to the JS file
+ * @param {Object} attrs additional optional attributes
+ */
+
```
jckautzmann, line 105: nit: remove the new line

PR #221, opened 2023-05-12
chore: add ios and macosx safari support for favicon

davidnuescheler:
this time for real :) 

https://favicon-update--helix-project-boilerplate--adobe.hlx.live/
vs.


rofe (approved):

rofe:
@davidnuescheler i still don't get a favicon in my safari tab...

davidnuescheler:
so after trying to figure out what's going on here, it turns out that safari in macos has an even different style favicon and it seems that short of supplying `.ico` and / or a `icon-mask` single color svg there is no way of getting this done.
the `touch-icon` works for iOS, and i don't know that it is valuable enough to be included in boilerplate just for safari on desktop.

davidnuescheler:
oh well, i tried it anyway... i am wondering though at this point, if we should remove the scripts completely and just go with the `favicon.ico`

rofe:
> i am wondering though at this point, if we should remove the scripts completely and just go with the `favicon.ico`

You're right, this can easily turn into a library with features of questionable use. Where would we stop? We could keep extending it to support Apple touch icons, Android icons, etc. Maybe that's something we should leave up to projects.

Let's keep it simple: by default, we assume a `/favicon.ico` is present (in the root). More sophisticated use cases could be described on a dedicated documentation page about favicons.

rofe (approved):

PR #225, opened 2023-05-26
Remove promise rejection from fetchPlaceholders

gszilvasy:
fetchPlaceholders method returns an object with the items set up in placeholders.xlsx.
Unsuccessful fetching / processing of the placeholders caused an unhandled exception, the issue is fixed in this PR. 
(The exception can also be handled in the functions calling the fetchPlaceholders function, but at the moment it is not handled)

Fix #<gh-issue-id>

Test URLs:
- Before: https://main--<repo>--<owner>.hlx.page/
- After: https://<branch>--<repo>--<owner>.hlx.page/

### DIFF
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -200,7 +200,7 @@ export async function fetchPlaceholders(prefix = 'default') {
   window.placeholders = window.placeholders || {};
   const loaded = window.placeholders[`${prefix}-loaded`];
   if (!loaded) {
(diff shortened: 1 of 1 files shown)


kptdobe (approved):
I think it makes sense. @fkakatie WDYT ?

kptdobe:
This is now fixed. See latest code version here: https://github.com/adobe/aem-lib/blob/main/src/placeholders.js

PR #228, opened 2023-06-15
feat: meta based csp

davidnuescheler:
setting a `CSP` that is only transported over the wire once and then cached on the client.
manages the `CSP` in an easy to read JSON file

vs.
https://meta-csp--helix-project-boilerplate--adobe.hlx.live/?rum=1

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -84,6 +84,7 @@ async function setCSP() {
   const meta = document.createElement('meta');
   meta.setAttribute('http-equiv', 'Content-Security-Policy');
   meta.setAttribute('content', policy);
+  document.addEventListener("securitypolicyviolation", (e) => sampleRUM("csperror", {source: `${e.documentURI}:${e.lineNumber}:${e.columnNumber}`, target: e.blockedURI }));
   document.head.appendChild(meta);
 }
 
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -84,7 +84,7 @@ async function setCSP() {
   const meta = document.createElement('meta');
   meta.setAttribute('http-equiv', 'Content-Security-Policy');
   meta.setAttribute('content', policy);
-  document.addEventListener("securitypolicyviolation", (e) => sampleRUM("csperror", {source: `${e.documentURI}:${e.lineNumber}:${e.columnNumber}`, target: e.blockedURI }));
+  document.addEventListener('securitypolicyviolation', (e) => sampleRUM('csperror', { source: `${e.documentURI}:${e.lineNumber}:${e.columnNumber}`, target: e.blockedURI }));
   document.head.appendChild(meta);
 }
 
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -73,10 +73,27 @@ async function loadEager(doc) {
 }
 
 /**
+ * sets the Content-Security-Policy meta tag to the document based on JSON file
+ */
+
+async function setCSP() {
+  const resp = await fetch(`${window.hlx.codeBasePath}/scripts/csp.json`);
+  const json = await resp.json();
+  const directives = Object.keys(json);
+  const policy = directives.map((directive) => `${directive} ${json[directive].join(' ')}`).join('; ');
+  const meta = document.createElement('meta');
+  meta.setAttribute('http-equiv', 'Content-Security-Policy');
+  meta.setAttribute('content', policy);
+  document.head.appendChild(meta);
+}
+
+/**
  * Loads everything that doesn't need to be delayed.
  * @param {Element} doc The container element
  */
 async function loadLazy(doc) {
+  await setCSP();
+
   const main = doc.querySelector('main');
   await loadBlocks(main);
 
--- a/scripts/csp.json
+++ b/scripts/csp.json
@@ -1,0 +1,13 @@
+{
+    "default-src": [
+        "'self'"
+    ],
+    "script-src": [
+        "'self'",
+        "https://rum.hlx.page/"
+    ],
+    "connect-src": [
+        "'self'",
+        "https://rum.hlx.page/"
+    ]
+}


scripts/scripts.js
```diff
@@ -72,11 +72,28 @@ async function loadEager(doc) {
...
+  meta.setAttribute('content', policy);
+  document.head.appendChild(meta);
+}
+
 /**
  * Loads everything that doesn't need to be delayed.
  * @param {Element} doc The container element
  */
 async function loadLazy(doc) {
+  await setCSP();
```
trieloff, line 95: I'd check if this is effective in blocking eagerly loaded resources that violate the CSP.

scripts/scripts.js
```diff
@@ -72,11 +72,28 @@ async function loadEager(doc) {
...
+
+async function setCSP() {
+  const resp = await fetch(`${window.hlx.codeBasePath}/scripts/csp.json`);
+  const json = await resp.json();
+  const directives = Object.keys(json);
+  const policy = directives.map((directive) => `${directive} ${json[directive].join(' ')}`).join('; ');
+  const meta = document.createElement('meta');
+  meta.setAttribute('http-equiv', 'Content-Security-Policy');
+  meta.setAttribute('content', policy);
+  document.head.appendChild(meta);
```
trieloff, line 87:
```suggestion
  document.addEventListener("securitypolicyviolation", (e) => sampleRUM("csperror", {source: `${e.documentURI}:${e.lineNumber}:${e.columnNumber}`, target: e.blockedURI }));
  document.head.appendChild(meta);
```

This would allow us to track CSP violations without overloading the endpoint as the same sampling applies as other RUM checkpoints.

trieloff (commented):
- track violations in RUM
- check for timing, violating resources might get loaded before CSP becomes effective

scripts/scripts.js
davidnuescheler: i am not sure that i am really that worried about the CSPs protecting block code...
to me this probably has to be in place before we execute untrusted code, eg. martech stack, so one could argue that this is in the wrong place altogether. so to a certain extent i think i am saying i don't even know when this should be loaded, so i am not really sure that an `await` really makes a difference here...

i am somewhat doubtful of the allowlist based approach altogether, so maybe a nonce based `strict-dynamic`  approach would be the right thing here, but i definitely don't know enough about the repercussions of that vis-a-vis the somewhat questionable benefit of CSPs around XSS attacks on franklin sites.

davidnuescheler: https://research.google/pubs/pub45542/

kptdobe:
For the record: `frame-ancestors` cannot be controlled via `meta` attribute (see https://www.w3.org/TR/CSP3/#directive-frame-ancestors), i.e. it won't work here. 
And it makes sense (browser would have already started to load the iframe once the directive comes in).

amol-anand:
Used this approach across multiple projects. Works great. I would like for this to be default in boilerplate to encourage folks to use CSP.

If someone adds something to the martech stack via a tag manager (Launch/GTM) or some iframe in the content via the Embed block, then they have to explicitly allow that in the CSP in code and create a PR which would run PSI checks against the site to ensure the new plugin is not impacting page performance. This will force better practices and ensure that it is easy to still maintain performance while also adding additional libraries to martech.

scripts/scripts.js
trieloff: Revisiting this, I think this is right, but only the first step. The next step should be @chicharr's lightweight tag management controlling the CSP

andreituicu: I think `setCSP` should be called as the first thing in `loadPage`, so the CSPs can offer protection as soon as possible. The additional request to fetch the configuration could be avoided in the critical path of FCP/LCP by inlining the configuration in the function. Something like:

```javascript
function setCSP() {
  const config = {
    "default-src": [
        "'self'"
    ],
    ...
  };

  const directives = Object.keys(config);
  const policy = directives.map((directive) => `${directive} ${config[directive].join(' ')}`).join('; ');
 ...
}
```

ramboz:
Is there a reason not to do it directly in the HTTP headers? It feels to me this would be more secure to begin with and you save yourself the additional resource in the LCP

scripts/scripts.js
```diff
@@ -72,11 +72,29 @@ async function loadEager(doc) {
   }
 }
 
+/**
+ * sets the Content-Security-Policy meta tag to the document based on JSON file
+ */
+
+async function setCSP() {
+  const resp = await fetch(`${window.hlx.codeBasePath}/scripts/csp.json`);
+  const json = await resp.json();
```
rofe, line 81: We could store the value of `csp.json` directly in `scripts.js` to avoid the additional request.

ramboz: Sure… but still not sure why we'd do that in JS code instead of just using headers since this doesn't change much usually. It's kind of a one-time setup

andreituicu: > but still not sure why we'd do that in JS code instead of just using headers since this doesn't change much usually. It's kind of a one-time setup

I completely agree that doing it via headers would be more secure. The disadvantage with JS is clear, you could always have an XSS kick in before the meta tag is set.

The only advantage that I see of the JS code is that it can be out of the box in the boilerplate and new projects would need to adapt it, so it creates awareness, that there's something they should do. We could even put a recommendation that they should move it to headers.

Alternatively, not sure how it could be done, but all new projects (to not break existing ones) could start with an out of the box CSP header, that needs to be deleted or modified.

ramboz: I'm more from the "secure by default" camp, and I'd prefer having it clearly in the pre go-live checklist than letting everyone assume that the JS approach will properly protect them all the time. I think giving the community the impression that you can fully handle this client-side is setting us up for escalations down the road when someone starts exploiting the loophole. I'd also argue that this is likely an IT or sec team operation as part of the production hardening that could likely be managed at the CDN level directly by experts, and not something I'd leave up to authors or "script kiddies".

My proposal would be to do:
1. a dedicated documentation page on this, and improve the CDN doc pages we have to cover this
2. creating some automation for pre go-live checks, like a PSI report but for the go-live topics we want them to check
3. clearly having those CSP set on aem.live and all reference projects we use
    - add it to the `headers` worksheet in the GDrive for the boilerplate
    - possibly create a dedicated sheet that mimics the JSON above and a computed value for the header if we really want to simplify the editing

kptdobe: The problem with the "experts" managing those at the CDN level is that they pile millions of rules, never clean it up and nobody knows what's in there. The header becomes so huge that it adds several KBs for every single request. We even had the case (several times) where it was so huge that request headers size was too large for our lambdas (more than 8KB!)...
Doing it client side would allow to cache the rules once and lighten all requests.

Why would it be less secure to do it client side ? Is this is a gut feeling or something backed with data ? If you are able to get rid of the rules if they are loaded client side, you can also get rid of the header...

andreituicu: > I'm more from the "secure by default" camp

I feel like we are all in the same camp. I think we are simply debating what is the most efficient and effective way to  make that happen, what is the best thing that we can put in place to make sure things don't remain just at documentation level.
 
I agree that the most secure way is to do it using http headers. 

> Why would it be less secure to do it client side ? Is this is a gut feeling or something backed with data ? If you are able to get rid of the rules if they are loaded client side, you can also get rid of the header...

It is more secure, because the header sets the CSP before any JS is executed.
If you set it on the client side and attacker might manage to do their XSS before the code that sets the CSP gets a chance to execute. Taking the example of the current implementation where `setCSP` is called in `loadLazy` an XSS could happen in an LCP block or something else running in `loadEager`, or if a customer has a script in `<head>` that does something before `scripts.js`.

> I'd prefer having it clearly in the pre go-live checklist

I agree there should be an entry in the pre go-live checklist/CDN configuration docs, but I feel like if during pre-golive it is the first time a CSP is discussed, it is too late. It means that all code has already been written and tested. Setting a CSP so late means everything needs to be retested to make sure all the dynamic scripts being loaded still work. 

I was thinking of the client side as a way to:
1. Create awareness of CSPs and starting the conversation early. (e.g. dynamically loading a 3rd party script is blocked by their browser, they adjust it) 
2. Building out the CSP during the development and the golive checklist/CDN config simply saying: now move that to a header (edge delivery header or CDN header).

> add it to the headers worksheet in the GDrive for the boilerplate

I agree this could work as an early signal for devs that they should start working on their CSP early and built it out during development, but only works if people actually copy that worksheet, whereas not using the boilerplate is probably less likely to happen.

> like a PSI report but for the go-live topics we want them to check

btw, the PSI report is already scanning and telling us all this.

Screenshot 2024-10-21 at 12 06 16

As David above, I'm wondering if a `nonce` + `strict-dynamic` wouldn't as an out of the box solution that works for majority of customers. https://content-security-policy.com/strict-dynamic/

davidnuescheler:
i think this can work in some situations, but i think this is too fickle for general consumption. i think having this referenced in block-party is probably the right place, and we probably need a more manageable CSP specific support in config service.

PR #231, opened 2023-06-28
Protect authors from invalid definitions.

### LINKED ISSUE #230: Invalid

bstopp:

Fix #230 

Test URLs:
- Before: https://main--franklin-demo--bstopp.hlx.page/
- After: https://bug-section-style-metadata--franklin-demo--bstopp.hlx.page/

### DIFF
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -347,7 +347,7 @@ export function decorateSections(main) {
       Object.keys(meta).forEach((key) => {
         if (key === 'style') {
           const styles = meta.style.split(',').map((style) => toClassName(style.trim()));
-          section.classList.add(...styles.filter(style => style));
+          section.classList.add(...styles.filter((style) => style));
         } else {
           section.dataset[toCamelCase(key)] = meta[key];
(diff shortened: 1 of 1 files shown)


davidnuescheler:
i think while we are at it we could also use something `.classList.add(...styles.filter(style => style))` which would improve on the .forEach(), and would keep the text transformation line (which is already a bit unreadable).

bstopp:
What about something like?

```
const styles = meta.style.split(',')
  .filter((style) => style.trim())
  .map((style) => toClassName(style));
styles.forEach((style) => section.classList.add(style));
```

bstopp:
What about?

```
          const styles = meta.style.split(',')
            .filter((style) => style.trim())
            .map((style) => toClassName(style.trim()));
          section.classList.add(...styles);
```

kptdobe:
@bstopp Closing this PR. This seems to be an important fix, could you bring it to https://github.com/adobe/aem-lib (+tests). Thanks a lot.

bstopp:
Moved to AEM Lib

PR #246, opened 2023-08-11
feat: add capability to patch block config at runtime

ramboz:
We've had several cases in the past where we needed to override the default block config logic when loading blocks:
- experimentations may load from a different folder name, or load a different js/css file in the same block folder
- themes that were loading alternate CSS files for block variants

To make maintaining those downstream projects easier, I propose to directly add an extension mechanism so we have fewer modifications to the `lib-franklin.js` file.

Test URLs:
- After:
  - https://extend-block-config--helix-project-boilerplate--ramboz.hlx.page/ (regular page not using the extension mechanism)
  - https://main--franklin-experience-decisioning--ramboz.hlx.page/experiments/control (page with experimentation using the extension mechanism)

Sample usage: https://github.com/ramboz/franklin-experience-decisioning/blob/main/scripts/experience-decisioning/index.js#L359-L417

### DIFF
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -430,7 +430,7 @@ export function buildBlock(blockName, content) {
 
 /**
  * Gets the configuration for the given glock, and also passes
- * the config to the `patchBlockConfig` methods in the plugins.
+ * the config throw all custom patching helpes added to the project.
  *
  * @param {Element} block The block element
  * @returns {object} The block config (blockName, cssPath and jsPath)
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -429,11 +429,11 @@ export function buildBlock(blockName, content) {
 }
 
 /**
- * Gets the configuration for the given glock, and also passes
- * the config throw all custom patching helpes added to the project.
+ * Gets the configuration for the given block, and also passes
+ * the config through all custom patching helpers added to the project.
  *
  * @param {Element} block The block element
- * @returns {object} The block config (blockName, cssPath and jsPath)
+ * @returns {Object} The block config (blockName, cssPath and jsPath)
  */
 function getBlockConfig(block) {
   const { blockName } = block.dataset;
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -440,8 +440,10 @@ function getBlockConfig(block) {
   const cssPath = `${window.hlx.codeBasePath}/blocks/${blockName}/${blockName}.css`;
   const jsPath = `${window.hlx.codeBasePath}/blocks/${blockName}/${blockName}.js`;
   const original = { blockName, cssPath, jsPath };
-  return window.hlx.patchBlockConfig.reduce(
-    (config, fn) => (typeof fn === 'function' ? fn(config, original) : config) || config,
+  return window.hlx.patchBlockConfig
+    .filter((fn) => typeof fn === 'function')
+    .reduce(
+    (config, fn) => fn(config, original),
     { blockName, cssPath, jsPath },
   );
 }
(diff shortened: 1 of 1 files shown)


scripts/lib-franklin.js
```diff
@@ -428,6 +428,24 @@ export function buildBlock(blockName, content) {
   return (blockEl);
 }
 
+/**
+ * Gets the configuration for the given glock, and also passes
+ * the config throw all custom patching helpes added to the project.
+ *
+ * @param {Element} block The block element
+ * @returns {object} The block config (blockName, cssPath and jsPath)
```
rofe, line 436:
```suggestion
 * Gets the configuration for the given block, and also passes
 * the config through all custom patching helpers added to the project.
 *
 * @param {Element} block The block element
 * @returns {Object} The block config (blockName, cssPath and jsPath)
```

rofe (changes requested):
LGTM, apart from a few jsdoc typos.

The sampe usage link looks wrong, or maybe the repo is private?

scripts/lib-franklin.js
ramboz: 🤦🏼 Thanks for catching those…
I don't think I ever made so many typos in just 3 lines 😓

ramboz:
@rofe Right, forgot I had the repo as private. I opened it up now.

rofe (approved):

scripts/lib-franklin.js
```diff
@@ -428,6 +428,24 @@ export function buildBlock(blockName, content) {
...
+ */
+function getBlockConfig(block) {
+  const { blockName } = block.dataset;
+  const cssPath = `${window.hlx.codeBasePath}/blocks/${blockName}/${blockName}.css`;
+  const jsPath = `${window.hlx.codeBasePath}/blocks/${blockName}/${blockName}.js`;
+  const original = { blockName, cssPath, jsPath };
+  return window.hlx.patchBlockConfig.reduce(
+    (config, fn) => (typeof fn === 'function' ? fn(config, original) : config) || config,
+    { blockName, cssPath, jsPath },
+  );
```
trieloff, line 446:
```suggestion
  return window.hlx.patchBlockConfig
    .filter((fn) => typeof fn === 'function')
    .reduce(
    (config, fn) => fn(config, original),
    { blockName, cssPath, jsPath },
  );
```

This makes it a bit more readable.

scripts/lib-franklin.js
```diff
@@ -428,6 +428,24 @@ export function buildBlock(blockName, content) {
...
+ * @returns {Object} The block config (blockName, cssPath and jsPath)
+ */
+function getBlockConfig(block) {
+  const { blockName } = block.dataset;
+  const cssPath = `${window.hlx.codeBasePath}/blocks/${blockName}/${blockName}.css`;
+  const jsPath = `${window.hlx.codeBasePath}/blocks/${blockName}/${blockName}.js`;
+  const original = { blockName, cssPath, jsPath };
+  return window.hlx.patchBlockConfig.reduce(
+    (config, fn) => (typeof fn === 'function' ? fn(config, original) : config) || config,
+    { blockName, cssPath, jsPath },
```
trieloff, line 445: Why not `original` to initialize the reduce?

trieloff (approved):

scripts/lib-franklin.js
ramboz: Was just to avoid someone modifying it inside the `fn` and potentially breaking the values.

ramboz:
@trieloff @rofe Once this is merged, do I need to formally document this somewhere?

rofe:
> @trieloff @rofe Once this is merged, do I need to formally document this somewhere?

Ideally yes. If there isn't already, there should probably be an experimentation page in the developer section of www.hlx.live.

ramboz:
@rofe This doesn't add experimentation, just an extension point for changing how block-related files (JS, CSS) are loaded.

I was wondering if extension points are documented somewhere and if I need to add this to it.

rofe:
> This doesn't add experimentation
I know, I just thought experimentation would be a concrete use case to document how this extension point can be used. 

>I was wondering if extension points are documented somewhere and if I need to add this to it.
I don't think so, but might be a good idea to start one.

PR #253, opened 2023-09-05
fix: revert to simple icons decoration and extract the spriting logic

ramboz:
Most customers directly use svg icons as they were explored by their preferred application (Illustrator, etc.), and that markup isn't compatible with the spriting logic in most cases. On top of that, the current spriting implementation has required frequent patching and isn't likely that stable yet. As such, we want to revert to basic icons decoration by default, and move the spriting to the block party instead until it is considered mature enough.

Test URLs:
- After: https://improve-icons--helix-project-boilerplate--adobe.hlx.page/

### DIFF
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -185,9 +185,9 @@ export async function decorateIcons(element) {
           return;
         }
         // Styled icons don't play nice with the sprite approach because of shadow dom isolation
-        // and same for internal references
+        // how colors are defined, and internal references handled
         const svg = await response.text();
-        if (svg.match(/(<style | class=|url\(#| xlink:href="#)/)) {
+        if (svg.match(/(<style | (class|fill|stroke)=|url\(#| xlink:href="#)/) && !svg.match(/"currentcolor"/i)) {
           ICONS_CACHE[iconName] = {
             styled: true,
             html: svg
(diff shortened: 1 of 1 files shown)


scripts/lib-franklin.js
```diff
@@ -184,26 +184,21 @@ export async function decorateIcons(element) {
...
-            styled: true,
-            html: svg
-              // rescope ids and references to avoid clashes across icons;
-              .replaceAll(/ id="([^"]+)"/g, (_, id) => ` id="${iconName}-${id}"`)
-              .replaceAll(/="url\(#([^)]+)\)"/g, (_, id) => `="url(#${iconName}-${id})"`)
-              .replaceAll(/ xlink:href="#([^"]+)"/g, (_, id) => ` xlink:href="#${iconName}-${id}"`),
-          };
-        } else {
+        // Icons using the "currentcolor" value are meant to inherit from the current
+        // page context and should be inlined
```
davidnuescheler, line 189: probably should say something like...

> Icons containing "currentcolor" will be used via symbols

scripts/lib-franklin.js
```diff
@@ -225,8 +220,10 @@ export async function decorateIcons(element) {
     const iconName = Array.from(span.classList).find((c) => c.startsWith('icon-')).substring(5);
     const parent = span.firstElementChild?.tagName === 'A' ? span.firstElementChild : span;
     // Styled icons need to be inlined as-is, while unstyled ones can leverage the sprite
```
davidnuescheler, line 222: probably should say:
> `<use>` sprite or `` for non-sprited icons

kptdobe:
We are currently exploding lib-franklin into its own library which includes adding tests... I think this refactor would be a perfect fit to add the corresponding tests to the method - See https://github.com/adobe/aem-lib/blob/main/src/decorate.js#L51. Maybe the PR should be moved there.

kptdobe:
@ramboz [https://github.com/adobe/aem-lib is ready (only need to adjust the documentation before release). Could you move your current work in this repo ? We will be able to write corresponding tests to all the icons handling edge cases ;)

davidnuescheler (approved):

ramboz:
@kptdobe Closing in favor of https://github.com/adobe/aem-lib/pull/15

davidnuescheler:
i think it would be great to also merge the PR here... i am when we are going to switch over to `aem-lib` codebase and i just wanted to update the documentation ahead of time...

kptdobe:
`aem-lib` is now in, the code is now in this repo.

PR #255, opened 2023-09-07
feat: use aem.js

kptdobe:
Use aem.js built from https://github.com/adobe/aem-lib.

Requires a rename of the repo from `helix-project-boilerplate` to `aem-boilerplate` right after merging.

Test URLs:
- After: https://aem-lib--helix-project-boilerplate--adobe.hlx.page/


trieloff (approved):

scripts/aem.js
```diff
@@ -129,211 +282,150 @@ export async function loadScript(src, attrs) {
...
-    const blockWrapper = block.parentElement;
-    blockWrapper.classList.add(`${shortBlockName}-wrapper`);
-    const section = block.closest('.section');
-    if (section) section.classList.add(`${shortBlockName}-container`);
-  }
+function decorateIcon(span, prefix = '') {
+  const iconName = Array.from(span.classList)
+    .find((c) => c.startsWith('icon-'))
+    .substring(5);
+  const img = document.createElement('img');
```
icaraps, line 405: @trieloff @kptdobe 

By using the `img` tag to load the SVG, you lose the ability to apply CSS to change its color and other properties. This is a common use case for small icons that for example should match the surrounding text. For example on https://blog.adobe.com/en/publish/2023/10/11/max-sneaks-2023, checkout the share icons. 

In some cases, using the `img` tag to load larger SVG illustrations makes sense as you don't want to increase the DOM size unnecessarily. 

So depending on the use case, ideally you should be able to decide whether you want to load the SVG with `<svg>` or with `` WDYT ?

icaraps: Also this change should have been done in a separate PR IMO. It's hidden behind the franklin to aem rename currently.

ramboz: @icaraps We've had many issues and edge cases with the SVG sprite approach we had introduced earlier and the decision was taken to go back to a simple decoration that non-experts can leverage out-of-the-box for GA.
(check https://github.com/adobe/aem-lib/pull/15 for more details)

Since styling icons essentially requires that the SVG file be modified from the OOTB export that apps like Illustrator give you, it is assumed to be a developer use case that can be done directly at the project level for now. And we are planning to introduce a plugin system in the near future that would allow you to pull in the sprite logic back on-demand.

icaraps: @ramboz Ok thanks for the details. Do you have recommendations on how to best migrate projects using svg icons to images instead ? For example, the icon selector Sidekick plugin built for Nokia uses svg icons https://main--nokia--hlxsites.hlx.page/tools/sidekick/library.html?plugin=icons. Do you foresee any issues if they switch to use images instead ?

ramboz: First, I'd try to use the new logic and see if it actually breaks anything. As far as I can see from the code, you weren't using the SVG sprite approach that we removed, and the change for you is just inline svg to SVG wrapped in an `` tag. It's likely the only impact would be the `currentcolor` inheritance from the spectrum themes, which might be easily searched/replaced globally if you don't use multiple variations of it.

If you do see regressions with the change, I'd suggest just copying over your current `decorateIcons` logic into the project code and leverage that instead of the central default one.

icaraps: @ramboz sounds good thanks !

PR #256, opened 2023-09-07
fix: race condition in decorateIcons

### LINKED ISSUE #217: multiple calls to decorateIcons leaves icons empty
When programatically calling `decorateIcons()`  more than once during the initial rendering ( e.g. from a block js that is present more than once in a page)  only the icon related to the first call will be printed,
the rest of the icons will be replaced with: https://github.com/adobe/helix-project-boilerplate/blob/main/scripts/lib-franklin.js#L189

The current hypothesis it concurrent access to ICONS_CACHE, the first call sets the value to true in:
https://github.com/adobe/helix-project-boilerplate/blob/main/scripts/lib-franklin.js#LL151C7-L151C18

and while the icon is loading for the first call to  `decorateIcons()`  , the second call to  `decorateIcons()`  does not evaluate https://github.com/adobe/helix-project-boilerplate/blob/main/scripts/lib-franklin.js#LL150C16-L150C16 but directly jumps to https://github.com/adobe/helix-project-boilerplate/blob/main/scripts/lib-franklin.js#L186.

karlpauls:
Fix #217 

When the same icon is requested several times (like a logo in the header and the footer) there is a race condition in the decorateIcons function that will cause only the first request to get the actual svg while the other requests will get empty icon tags. 

This PR fixes this by replacing the current logic that is based on just putting a boolean in the cache while the icon gets loaded with a promise that is put into the cache (similar to what happens in fetchPlaceholders). Subsequently, that promise is returned from the cache for all requesters - hence, they all get the result.

We used this patch in our project and haven't had any issues since then. 

Test URLs:
- After: https://217--helix-project-boilerplate--karlpauls.hlx.live/

### DIFF
--- a/scripts/lib-franklin.js
+++ b/scripts/lib-franklin.js
@@ -174,43 +174,48 @@ export async function decorateIcons(element) {
 
   // Download all new icons
   const icons = [...element.querySelectorAll('span.icon')];
-  await Promise.all(icons.map(async (span) => {
+  await Promise.all(icons.map((span) => {
     const iconName = Array.from(span.classList).find((c) => c.startsWith('icon-')).substring(5);
-    if (!ICONS_CACHE[iconName]) {
-      ICONS_CACHE[iconName] = true;
-      try {
-        const response = await fetch(`${window.hlx.codeBasePath}/icons/${iconName}.svg`);
-        if (!response.ok) {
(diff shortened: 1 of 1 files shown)


kptdobe:
Fixed in https://github.com/adobe/aem-lib/pull/12

PR #264, opened 2023-09-29
chore: add fragments and use in header / footer

davidnuescheler:
starting the move of fragments to boilerplate and using fragments to load header and footer.

- Before: 
https://main--aem-boilerplate--adobe.hlx.live/
https://main--aem-boilerplate--adobe.hlx.live/drafts/uncled/fragment-test

- After: 
https://fragments--aem-boilerplate--adobe.hlx.live/
https://fragments--aem-boilerplate--adobe.hlx.live/drafts/uncled/fragment-test


blocks/header/header.js
```diff
@@ -90,56 +91,58 @@ function toggleMenu(nav, navSections, forceExpanded = null) {
  * @param {Element} block The header block element
  */
 export default async function decorate(block) {
-  // fetch nav content
+  // load nav as fragment
   const navMeta = getMetadata('nav');
   const navPath = navMeta ? new URL(navMeta).pathname : '/nav';
-  const resp = await fetch(`${navPath}.plain.html`);
+  const fragment = await loadFragment(`${navPath}`);
```
synox, line 97: can be used without template string

blocks/footer/footer.js
```diff
@@ -1,25 +1,21 @@
...
-  const cfg = readBlockConfig(block);
+  const footerMeta = getMetadata('footer');
   block.textContent = '';
 
-  // fetch footer content
-  const footerPath = cfg.footer || '/footer';
-  const resp = await fetch(`${footerPath}.plain.html`, window.location.pathname.endsWith('/footer') ? { cache: 'reload' } : {});
+  // load footer fragment
+  const footerPath = footerMeta.footer || '/footer';
+  const fragment = await loadFragment(`${footerPath}`);
```
rofe:
```suggestion
  const fragment = await loadFragment(footerPath);
```

blocks/header/header.js
rofe:
```suggestion
  const fragment = await loadFragment(navPath);
```

rofe (approved):

blocks/fragment/fragment.js
```diff
@@ -0,0 +1,55 @@
+/*
+ * Fragment Block
+ * Include content from one Helix page in another.
```
fkakatie: Probably worth rethinking the caption.

fkakatie (approved):

blocks/fragment/fragment.js
davidnuescheler: fixed it ;)

blocks/footer/footer.js
```diff
@@ -1,25 +1,21 @@
...
 export default async function decorate(block) {
-  const cfg = readBlockConfig(block);
+  const footerMeta = getMetadata('footer');
   block.textContent = '';
 
-  // fetch footer content
-  const footerPath = cfg.footer || '/footer';
-  const resp = await fetch(`${footerPath}.plain.html`, window.location.pathname.endsWith('/footer') ? { cache: 'reload' } : {});
+  // load footer fragment
+  const footerPath = footerMeta.footer || '/footer';
```
arumsey, line 13: @davidnuescheler I think this line should just be 

```
const footerPath = footerMeta || '/footer';
```

since switching to `getMetadata` instead of `readBlockConfig`. IIUC, `getMetadata` only returns a string so `footerMeta.footer` will always be undefined.

Is my assumption correct?

PR #274, opened 2023-10-25
Add reference to codeBasePath in header/footer

### LINKED ISSUE #273: Add codeBasePath reference in header and footer blocks
Reference of `${window.hlx.codeBasePath}` to be added by default in header and footer blocks.
This will be helpful for projects dealing with subpaths for site URL

asthabh23:

Fix #273 

Test URLs:
- After: https://issue-273--aem-boilerplate--adobe.hlx.page/

### DIFF
--- a/blocks/footer/footer.js
+++ b/blocks/footer/footer.js
@@ -10,7 +10,7 @@ export default async function decorate(block) {
 
   // fetch footer content
   const footerPath = cfg.footer || '/footer';
-  const resp = await fetch(`${footerPath}.plain.html`, window.location.pathname.endsWith('/footer') ? { cache: 'reload' } : {});
+  const resp = await fetch(`${window.hlx.codeBasePath}${footerPath}.plain.html`, window.location.pathname.endsWith('/footer') ? { cache: 'reload' } : {});
 
   if (resp.ok) {
     const html = await resp.text();
--- a/blocks/header/header.js
+++ b/blocks/header/header.js
@@ -93,7 +93,7 @@ export default async function decorate(block) {
   // fetch nav content
   const navMeta = getMetadata('nav');
   const navPath = navMeta ? new URL(navMeta).pathname : '/nav';
-  const resp = await fetch(`${navPath}.plain.html`);
+  const resp = await fetch(`${window.hlx.codeBasePath}${navPath}.plain.html`);
 
   if (resp.ok) {
     const html = await resp.text();


ramboz:
@asthabh23 I'm not sure I agree with the change. I think we are mixing content and code concerns here.
My understanding is that `window.hlx.codeBasePath` would point to the root of the code, relative to the root of the server (i.e. if code is installed in `/aem/…` this lets us dynamically adjust all references in the code accordingly)
Content on the other hand could still be coming from the server root, and thus prefixing the fragments, to me, seems wrong here. I'd argue this is something to be defined at the project level, possibly with a new `window.hlx.contentBasePath`, but I wouldn't apply this to the boilerplate globally.

@rofe @kptdobe @davidnuescheler What are your thoughts on this?

kptdobe:
As a first comment, the problem @asthabh23 is trying to solve is not clearly described in the issue. That would help to understand what you are trying to achieve.

I agree this is partially wrong. Several things: 
1. reading the original code, the nav and footer do not work the same - which is probably something to harmonise
2. the nav accepts a url: the url should be correct, no need to prepend anything
3. the footer accepts a path in the block config: the path should be correct, no need to prepend anything (this probably need to be a url for alignment)
4. the defaults (`/nav` and `/footer`) which I agree could be wrong if you put the corresponding pages in a subfolder

As @ramboz says, this is not `code path` related. But it is not completely `content path` related: you could have a custom context path, i.e. `/express` for all your content and still want to store your nav / footer under `/express/global` or something like that.

asthabh23:
@ramboz @kptdobe , thanks for your comments.
For a better context, this change is required for uniformity in projects even when we need to migrate subpath for a site.
To start with, we have codeBasePath null as here : https://github.com/adobe/aem-boilerplate/blob/main/scripts/aem.js#L137,
For projects where we need the code also in a directory, something like we have in express-website: https://github.com/adobe/express-website, we need to override this value in scripts.js so it actually adapts to that folder.

With that in place, we need to add this reference in `nav.js`, `footer.js` so it actually picks up the right reference.
@rofe suggested this needs to be there as default in boilerplate too, so we don't have to manually update in blocks.

rofe:
I agree that we should not mix code and content prefixes, since they can be different in theory - even though they're often identical in the real world (e.g. a CDN is routing all requests starting with `/foo` to Edge Delivery, so both code and content URLs must start with that). I like the proposal to add a `window.hlx.contentBasePath` (which can default to `window.hlx.contentbasePath`), but adding that would be a discussion to be had in https://github.com/adobe/aem-lib before blocks in boilerplate or block collection can use it. 

In any case, a content prefix should only be applied where default values are being used, and the URL is not authored.

PR #280, opened 2023-11-01
Add alt text support to decorateIcon

### LINKED ISSUE #279: Update to Icon SVG Handling created Accessibility regression
### Expected Behaviour

Icons should contain some form of `alt` text for the image tag.

### Actual Behaviour

No `alt` text exists, and thus Accessibility issue is raised on PSI check.

#### Steps to Reproduce

Load any demo page. [Here is a PSI check](https://pagespeed.web.dev/analysis/https-main--aem-eds-demo--bstopp-hlx-page/og2eede810?form_factor=mobile) of a page with multiple icon references.

bstopp:
Default the Alt Text to the iconName as referenced in markdown. 

Also support i18n lookup based on `icon-<icon-name>` key.

Fix #279 

Test URLs:
- After: https://feat-icon-alt-text--aem-eds-demo--bstopp.hlx.page/

### DIFF
--- a/scripts/aem.js
+++ b/scripts/aem.js
@@ -399,7 +399,8 @@ function decorateButtons(element) {
  * @param {string} [prefix] prefix to be added to icon the src
  */
 async function decorateIcon(span, prefix = '') {
-  const langPrefix = document.documentElement.lang === 'en' ? '' : document.documentElement.lang
+  const langPrefix = document.documentElement.lang === 'en' ? '' : document.documentElement.lang;
+  // eslint-disable-next-line no-use-before-define
   const placeholders = await fetchPlaceholders(langPrefix);
   const iconName = Array.from(span.classList)
     .find((c) => c.startsWith('icon-'))
@@ -408,7 +409,7 @@ async function decorateIcon(span, prefix = '') {
   const img = document.createElement('img');
   img.dataset.iconName = iconName;
   img.src = `${window.hlx.codeBasePath}${prefix}/icons/${iconName}.svg`;
-  img.alt =  placeholders[toCamelCase(`icon-${iconName}`)] || iconName;
+  img.alt = placeholders[toCamelCase(`icon-${iconName}`)] || iconName;
   img.loading = 'lazy';
   span.append(img);
 }
--- a/scripts/aem.js
+++ b/scripts/aem.js
@@ -395,16 +395,20 @@ function decorateButtons(element) {
 
 /**
  * Add <img> for icon, prefixed with codeBasePath and optional prefix.
- * @param {span} [element] span element with icon classes
+ * @param {Element} [span] span element with icon classes
  * @param {string} [prefix] prefix to be added to icon the src
  */
-function decorateIcon(span, prefix = '') {
+async function decorateIcon(span, prefix = '') {
+  const langPrefix = document.documentElement.lang === 'en' ? '' : document.documentElement.lang
+  const placeholders = await fetchPlaceholders(langPrefix);
   const iconName = Array.from(span.classList)
     .find((c) => c.startsWith('icon-'))
     .substring(5);
+
   const img = document.createElement('img');
   img.dataset.iconName = iconName;
   img.src = `${window.hlx.codeBasePath}${prefix}/icons/${iconName}.svg`;
+  img.alt =  placeholders[toCamelCase(`icon-${iconName}`)] || iconName;
   img.loading = 'lazy';
   span.append(img);
 }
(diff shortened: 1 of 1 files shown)


bstopp:
I don't know how to provide a valid PSI after URL. I can push a branch to the upstream repo, however I do not have rights to the Content source to make the change there to reflect tests.

Here's the [PSI check](https://pagespeed.web.dev/analysis/https-feat-icon-alt-text--aem-eds-demo--bstopp-hlx-page/6xrvqxfad7?form_factor=mobile) of my demo page

davidnuescheler:
i don't think that the decorate method (as it is called very early) should have a dependency on placeholders as that may add (variable and possibly considerable) weight to before LCP. 

i think icons and alt texts and title / description inside the SVGs are an interesting quandary where i am not totally sure what's the correct solution, but i think that they don't need to be added before the icons are loaded (via lazy or intersection observer).

i think @ramboz has a PR (i forgot where) to enable spriting (and title / alt handling).

bstopp:
> i think @ramboz has a PR (i forgot where) to enable spriting (and title / alt handling).

All sprites were removed during the renaming process. to `aem.js` Unless there's a new proposal out there to re-add sprite logic? (I haven't seen any new ones).

Even if the SVG has title/alt text inside the file itself, i'm not sure that solves this issue - Since the `img` tag itself doesn't have this information. I added titles to the _LinkedIn_ svg on my test page, and it didn't resolve the issue for that case ([PSI Report for non-fix branch](https://pagespeed.web.dev/analysis/https-main--aem-eds-demo--bstopp-hlx-page/o4o3z5a5mo?form_factor=mobile)) It seems as though if the SVG is inlined it meets the Accessibility requirements, but not if it is a img src.

@amol-anand indicated this might be a conern. I will remove the placeholders, but there should always be a default - it can just be the `iconName` as provided by the Author. I'll provide an example in the Block Party on how to post-process icons to provide a more meaningful alt description using placeholders.

bstopp:
[Updated PSI Report](https://pagespeed.web.dev/analysis/https-feat-icon-alt-text--aem-eds-demo--bstopp-hlx-page/p6rctbg6pc?form_factor=mobile)

scripts/aem.js
```diff
@@ -395,16 +395,18 @@ function decorateButtons(element) {
 
 /**
  * Add <img> for icon, prefixed with codeBasePath and optional prefix.
- * @param {span} [element] span element with icon classes
+ * @param {Element} [span] span element with icon classes
  * @param {string} [prefix] prefix to be added to icon the src
  */
 function decorateIcon(span, prefix = '') {
```
ramboz: I think we need to make this configurable. It only makes sense to add the alt attribute if we have an actual valuable description to expose via screen readers. If the icons are only used for decoration purposes, then it's usually better to just annotate them with `role="presentation" alt=" "`.

I'd rather propose something like:
```suggestion
function decorateIcon(span, prefix = '', alt = '') {
```

scripts/aem.js
```diff
@@ -395,16 +395,18 @@ function decorateButtons(element) {
...
   const iconName = Array.from(span.classList)
     .find((c) => c.startsWith('icon-'))
     .substring(5);
+
   const img = document.createElement('img');
   img.dataset.iconName = iconName;
   img.src = `${window.hlx.codeBasePath}${prefix}/icons/${iconName}.svg`;
+  img.alt = iconName;
```
ramboz:
```suggestion
  if (alt && alt.trim()) {
    img.alt = alt;
  } else {
    img.alt = ' ';
    img.setAttribute('role', 'presentation');
  }
```

ramboz:
Right, I submitted the old spriting to the block party, but haven't heard back from the team yet.

You can find the PR here: https://github.com/ramboz/aem-boilerplate/pull/12

scripts/aem.js
bstopp: I did commit this, but my only concern with this is - there's no value coming from the `decorateIcons` wrapping iteration. This function is not exported, so where should the caller source this value?

ramboz: You'd have to export the single `decorateIcon` as well and then explicitly instrument in the blocks that require proper `alt` text

rofe:
@bstopp i think this PR needs to be moved to `aem-lib` since `aem.js` is released to this repo from there.

bstopp:
Didn't realize that. i'll move it this weekend.

fkakatie:
> Didn't realize that. i'll move it this weekend.

@bstopp, please take a look here when you get a chance: https://github.com/adobe/aem-lib/pull/26

bstopp:
Closing as moved to aem-lib

PR #303, opened 2023-12-07
refactor: use window.location as base for url construction

fkakatie:
Replaces instances of creating URL objects with a single argument with the two-argument form, using `window.location` as the base URL. Ensures that relative URLs are resolved against the current browser location for consistent handling.

Test URLs:
- After: https://relative-paths--aem-boilerplate--adobe.hlx.live/

### DIFF
--- a/blocks/footer/footer.js
+++ b/blocks/footer/footer.js
@@ -10,7 +10,7 @@ export default async function decorate(block) {
   block.textContent = '';
 
   // load footer fragment
-  const footerPath = footerMeta.footer || '/footer';
+  const footerPath = footerMeta ? new URL(footerMeta, window.location).pathname : '/footer';
   const fragment = await loadFragment(footerPath);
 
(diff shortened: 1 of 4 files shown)


kptdobe (approved):

blocks/footer/footer.js
```diff
@@ -10,7 +10,7 @@ export default async function decorate(block) {
   block.textContent = '';
 
   // load footer fragment
-  const footerPath = footerMeta.footer || '/footer';
+  const footerPath = footerMeta ? new URL(footerMeta, window.location).pathname : '/footer';
```
arumsey, line 13: This also fixes the bug (https://github.com/adobe/aem-boilerplate/pull/264) with `footerMeta` now being a string and not an object so `footerMeta.footer` can never be valid.

buuhuu:
@fkakatie this has now been done with #346

PR #308, opened 2023-12-09
feat: section loader

davidnuescheler:
Test URLs:
- After: https://section-loader--aem-boilerplate--adobe.hlx.live/

this is just a test, do not merge.


davidnuescheler:
this removes complexity and makes LCP guessing more flexible.

davidnuescheler:
image

davidnuescheler:
image

scripts/aem.js
```diff
@@ -675,6 +631,39 @@ async function waitForLCP(lcpBlocks) {
...
+ * Loads all blocks in a section.
+ * @param {Element} section The section element
+ */
+
+async function loadSection(section, loadCallback) {
+  const status = section.dataset.sectionStatus;
+  if (status === 'initialized') {
+    section.dataset.sectionStatus = 'loading';
+    const blocks = [...section.querySelectorAll('div.block')];
+    for (let i = 0; i < blocks.length; i += 1) {
```
shsteimer, line 704: @davidnuescheler this may be counter-intuitive, but if we go this route, would it make sense to actually load the blocks within a section in reverse order.

My thinking:

I've had projects where you end up with nested blocks via auto-blocking. when these blocks are loaded sequentially, the outer block is loaded before the inner block, meaning that the outer block decorate function cannot make decisions based on the content of the inner blocks (because it isn't yet fully formed). There are ways around this on a project basis of course, but it can get messy.

we'd never want to load the blocks for a full page in reverse order as it would delay appearance of top of the page, but since no section gets reveled until all of it's blocks are loaded, maybe within each section it would be safe/performant to do so, allowing nested blocks to be loaded before their `"parent"` blocks, and thus resolving the issue.

(or maybe this is only a problem on a small enough percentage of projects to just not worry about it. feel free to just ignore this)

davidnuescheler: that's very interesting indeed, i definitely left nested blocks via auto blocking kind of out of the equation... i think you are right that reverse order would yield better results in the sense that it may simplify that (special) case... TBH though, i have not been in a situation where i needed to make decisions in a block based on the nested block (in JS that is).
as you point out, i agree that it would be counter intuitive, and based on the sequence of https://github.com/adobe/helix-home/blob/main/manifesto.md#intuitive-simple--consistent-p i would probably at this stage defer to a later stage to incorporate this. personally, i think the nesting probably should be addressed with a more recursive approach at some point... at least that would be most intuitive to me.

scripts/aem.js
```diff
@@ -675,6 +631,39 @@ async function waitForLCP(lcpBlocks) {
...
+}
+
+/**
+ * Loads all sections.
+ * @param {Element} element The parent element of sections to load
+ */
+
+async function loadSections(element) {
+  const sections = [...element.querySelectorAll('div.section')];
+  for (let i = 0; i < sections.length; i += 1) {
```
auniverseaway, line 721:
```
for (const [idx, section] of sections.entries()) {
  await loadSection(section);
}
```

auniverseaway:
I'm just seeing this. I think this PR is great. This is much closer to what we did on Milo.

https://github.com/da-sites/nexter/blob/main/nx/scripts/nexter.js#L125-L136

One thing we also did in Milo is that we do not allow a block to "pollute up" their class names into sections. We saw a lot of CSS written in Express due to competing section styles because of block names being attached on sections.

Screenshot 2024-03-16 at 1 23 32 PM

**Our guiding principles on this topic:**

1. A block should never change the behavior of a section.
2. A section should never impact the internals of a block.

scripts/aem.js
```diff
@@ -675,6 +631,39 @@ async function waitForLCP(lcpBlocks) {
...
+
+async function loadSection(section, loadCallback) {
+  const status = section.dataset.sectionStatus;
+  if (status === 'initialized') {
+    section.dataset.sectionStatus = 'loading';
+    const blocks = [...section.querySelectorAll('div.block')];
+    for (let i = 0; i < blocks.length; i += 1) {
+      // eslint-disable-next-line no-await-in-loop
+      await loadBlock(blocks[i]);
+    }
```
kptdobe, line 707:
```suggestion
    const promises = blocks.map(block => loadBlock(block));
    await Promise.all(promises);
```
The `for` loop still makes the block loading sequential, i.e. block1, block2... blockN. If block 1 requires something remote and slow, it "blocks" all the other blocks processing. This can be avoided by parallelizing the block loading sequence.
The section is hidden anyway, no visible impact.

kptdobe (changes requested):
I have tested the loading sequence in the context of the bulk project (not in production -yet). It helps solving several issues, like the case where the LCP image is part of block 2: no way to make it work with current implementation.
I even think the current implementation (`waitForLCP`) is incorrect: we wait for the first image of the first block but it only gets displayed in the lazy phase when the first section is processed. We ned to change that asap.

scripts/aem.js
kptdobe: It seems I have proven in the past that this does not bring any real performance benefit AND makes the block code potentially more complicated (for example, if 2 blocks uses the same resource - like metadata.json, you need a singleton to make sure it is loaded only once...). This makes this change unnecessary.

kptdobe (approved):

kptdobe:
Working on the aem-lib version of this PR.

scripts/aem.js
kptdobe: Linting does not like this approach...
image

kptdobe:
I have created the PR in aem-lib to match all the changes required in aem.js: https://github.com/adobe/aem-lib/pull/80

I have aligned the aem.js of this PR with the build result of the aem-lib PR for testing purposes. Once the aem-lib PR is merged, this one can be merged without the aem.js changes.

kptdobe:
Replaces https://github.com/adobe/aem-boilerplate/pull/382.

PR #346, opened 2024-04-09
fix: header/footer fragment loading (#13)

buuhuu:
This PR fixes the broken footer loading from a given `footer` field in the page metadata. 

It also addresses the use case where the `footer` or `nav` are configured as paths only instead of a link, for example when using bulk metadata it seems reasonable to configure `footer` and `nav` like

| Path | Nav | Footer |
| ----- | --- | ---- |
| /** | /nav | /footer |
| /de** | /de/nav | /de/footer |

- After: https://fix-header-footer-loading--aem-boilerplate--buuhuu.hlx.live/

### DIFF
--- a/blocks/footer/footer.js
+++ b/blocks/footer/footer.js
@@ -10,7 +10,7 @@ export default async function decorate(block) {
   block.textContent = '';
 
   // load footer fragment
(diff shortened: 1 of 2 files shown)


fkakatie (approved):
🎉  Great catch.

PR #353, opened 2024-04-19
Add editor config

sippsolutions:
Add editor config

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+[*.js]
+indent_size = 2
+
+[*.css]
+indent_size = 4


kptdobe:
@sippsolutions Why are you trying to patch the aem-boilerplate ?

sippsolutions:
@kptdobe because in my opinion a `.editorconfig` is missing that should be equivalent in all projects.
I intentionally wanted to improve the boilerplate :) Can you reopen the PR?

kptdobe:
`.editorconfig` is a VS Code specific config file which has nothing to do with the boilerplate code repo.
What about other IDE specific config files that other dev might be using ? It would quickly be the mess.

kptdobe (approved):

kptdobe:
ok, thanks. I thought `linting` would be enough but since it requires a `npm install` first, I have seen so many projects where linting is not respected and fails. Let see if `.editorconfig` helps.

PR #362, opened 2024-05-24
chore: unify font

fkakatie:
* update font styles (size, weight, and line height) to align with s2 design

⚠️ potentially controversial: added a new font for the `heading-font-family` variable. boilerplate uses Roboto, which is already a departure from s2. most projects have multiple fonts, so including an example on how to properly configure those might be useful.

Test URLs:
- After: https://unify-font--aem-boilerplate--adobe.hlx.live/


rofe:
Is S2 compliance really a requirement for the boilerplate?

fkakatie:
> Is S2 compliance really a requirement for the boilerplate?

Not at all. I'm only pulling values from it in an attempt to set some sort of visual consistency across boilerplate > block collection.

rofe (approved):

PR #368, opened 2024-06-05
chore: remove unused dep

kptdobe:
Not sure why those 2 are still here...


fkakatie (approved):
good catch, thank you 🎉

PR #371, opened 2024-06-19
feat: use rum js v2

kptdobe:
Migration to rum js v2. A lot of checkpoints are not required anymore.

Depends on https://github.com/adobe/aem-lib/pull/76

The `aem.js` in this PR would come from https://github.com/adobe/aem-lib/pull/76, it does not need to be committed, but it is here to be able to test.

Test page: https://migration-rum-js-v2--aem-boilerplate--adobe.aem.live/?rum=on

### DIFF
--- a/404.html
+++ b/404.html
@@ -26,7 +26,7 @@
           btnContainer.append(backBtn);
         }
       }
-      sampleRUM('404', { source: document.referrer, target: window.location.href });
+      sampleRUM('404', { source: document.referrer });
     });
   </script>
   <link rel="stylesheet" href="/styles/styles.css">
--- a/404.html
+++ b/404.html
@@ -11,8 +11,6 @@
   <meta property="og:title" content="Page not found">
   <script src="/scripts/scripts.js" type="module" crossorigin="use-credentials"></script>
   <script type="module">
-    import { sampleRUM } from '/scripts/aem.js';
-
     window.addEventListener('load', () => {
       if (document.referrer) {
         const { origin, pathname } = new URL(document.referrer);
@@ -26,8 +24,11 @@
           btnContainer.append(backBtn);
         }
       }
-      sampleRUM('404', { source: document.referrer });
     });
+  </script>
+  <script type="module">
+    import { sampleRUM } from '/scripts/aem.js';
+    sampleRUM('404', { source: document.referrer });
   </script>
   <link rel="stylesheet" href="/styles/styles.css">
   <style>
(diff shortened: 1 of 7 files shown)


404.html
```diff
@@ -26,7 +26,7 @@
           btnContainer.append(backBtn);
         }
       }
-      sampleRUM('404', { source: document.referrer });
+      sampleRUM('404', { source: document.referrer, target: window.location.href });
```
trieloff:
```suggestion
      sampleRUM('404', { source: document.referrer });
```

trieloff: The `target` is redundant with the URL, no need to send it twice.

scripts/aem.js
```diff
@@ -124,17 +112,8 @@ function setup() {
...
-
-  window.addEventListener('unhandledrejection', (event) => {
-    sampleRUM('error', { source: event.reason.sourceURL, target: event.reason.line });
-  });
-
-  window.addEventListener('error', (event) => {
-    sampleRUM('error', { source: event.filename, target: event.lineno });
-  });
+  sampleRUM();
+  window.addEventListener('hlx:section:loaded', sampleRUM.enhance, { once: true });
```
trieloff:
```suggestion
  window.addEventListener('aem:section:loaded', sampleRUM.enhance, { once: true });
```

scripts/aem.js
```diff
@@ -533,6 +512,7 @@ function updateSectionsStatus(main) {
       } else {
         section.dataset.sectionStatus = 'loaded';
         section.style.display = null;
+        window.dispatchEvent(new CustomEvent('hlx:section:loaded', { detail: { section } }));
```
trieloff:
```suggestion
        window.dispatchEvent(new CustomEvent('aem:section:loaded', { detail: { section } }));
```

trieloff (approved):

trieloff:
No re-review required.

404.html
kptdobe: Humm.. @chicharr didn't we check we get a referer and the current page url... ?

scripts/aem.js
kptdobe: To be discussed, there is the idea that `hlx` namespace could be "internal" (i.e. do not rely on it, can change), `aem` being public (i.e. backward comp)...

404.html
kptdobe: I just checked. `document.referrer` is there when there is a referer and yes, of course,  `target` is a duplicated of  `cdn.url` we are tracking automatically.

PR #376, opened 2024-06-27
chore: change from `delayed` flow to include consent

davidnuescheler:
changing the flow of loading marketing tech based on a conditions / events such as consent

https://marketing-tech--aem-boilerplate--adobe.aem.page/

PR #408, opened 2024-10-02
chore: use events as trigger to load delayed.js

davidnuescheler:
in some situations it is desirable to load delayed on an earlier trigger than a specified timeout.

https://interaction-trigger-delayed--aem-boilerplate--adobe.aem.page/

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -120,7 +120,9 @@ async function loadLazy(doc) {
 function loadDelayed() {
   // eslint-disable-next-line import/no-cycle
   window.setTimeout(() => import('./delayed.js'), 3000);
-  // load anything that can be postponed to the latest here
+
+  // add other event-listeners that should trigger loading of delayed before timeout
+  document.addEventListener('scroll', () => import('./delayed.js'), { once: true });
 }
 
 async function loadPage() {
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -138,11 +138,25 @@ async function loadLazy(doc) {
  * without impacting the user experience.
  */
 function loadDelayed() {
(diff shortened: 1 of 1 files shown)


davidnuescheler:
this is likely to be obsoleted when https://github.com/adobe/aem-boilerplate/pull/376 is merged

freshjones:
@davidnuescheler this looks/works great, just have a couple questions:
1) since it seems like boilerplate is headed more in the direction of https://github.com/adobe/aem-boilerplate/pull/376 should we wait for that to get merged before considering this approach for our sites? I think our goal is to try and remain as aligned w the Adobe boilerplate approach as much as we can.
2) if we do use this approach is there any value in adding event listeners for additional events beyond scroll? e.g.
'mousemove',
    'touchstart',
    'keydown',
    'click',
    'touchmove',
    
    thanks!

davidnuescheler:
@freshjones thanks a lot for the input... very helpful.

(1) i think that's definitely true that these are competing and i don't know what the merge schedule of either of them are, but it is definitely a good idea to keep an eye on them. even after we merge #376 i would expect people to still use the traditional loading sequence as well, so this is truly just a recommended template.

(2) i think that projects are welcome to add other events, but i found the list somewhat arbitrary, so i didn't want to include and maintain a large set of event here, but i would expect projects to be able add more events based on their specific needs and uses.

fkakatie:
@davidnuescheler - FYI: I just updated `loadDelayed()` with multiple interaction triggers and duplicate import guards. And updated to `main`.

#376 removes `loadDelayed()` entirely for consent flow, which conflicts (obviously, discussed above). Worth considering whether consent should also wait for interaction vs. immediate load. And which PR to move forward with, if any.

PR #416, opened 2024-10-16
feat: Configure Mozilla's Eslint Plugin eslint-plugin-no-unsanitized + DOMPurify.sanitize + escapeHtml

andreituicu:
Building on top of https://github.com/adobe/aem-boilerplate/pull/415

Test urls:
* Before: https://main--aem-boilerplate--adobe.aem.live/
* After: https://xss-lint-dompurify--helix-project-boilerplate--andreituicu.aem.live/


fkakatie:
results of aem-psi-check would be necessary here.

andreituicu:
@fkakatie completely agree. It looks like there's a problem with my fork, because when I access https://xss-lint-dompurify--helix-project-boilerplate--andreituicu.aem.live the head doesn't have any scripts/css... Could I possibly get write permissions to the boilerplate? That way I can create the branches directly in the project and recreate the PRs including PSI values.

blocks/fragment/fragment.js
```diff
@@ -22,7 +22,8 @@ export async function loadFragment(path) {
     const resp = await fetch(`${path}.plain.html`);
     if (resp.ok) {
       const main = document.createElement('main');
-      main.innerHTML = await resp.text();
+
+      main.innerHTML = DOMPurify.sanitize(await resp.text());
```
rofe, line 26: Note that the original `fragment` block source is in the [block-collection](https://github.com/adobe/aem-block-collection/tree/main/blocks/fragment) and should probably be fixed there first, then updated here.

head.html
```diff
@@ -1,4 +1,5 @@
 <meta name="viewport" content="width=device-width, initial-scale=1"/>
+<script src="scripts/purify.min.js" type="text/javascript"></script>
```
rofe: I think we should dynamically import `purify.min.js` within the `fragment` block to avoid loading it on all pages that don't have any fragments.

scripts/aem.js
```diff
@@ -537,7 +537,7 @@ function buildBlock(blockName, content) {
       vals.forEach((val) => {
         if (val) {
           if (typeof val === 'string') {
-            colEl.innerHTML += val;
+            colEl.textContent += val;
```
rofe, line 540: Note that the `aem.js` source is [here](https://github.com/adobe/aem-lib) and the build result gets copied to the boilerplate.

scripts/scripts.js
```diff
@@ -14,6 +14,27 @@ import {
...
+ */
+export function escapeHtml(unsafe) {
+  return unsafe
+    .replaceAll('&', '&amp;')
+    .replaceAll('<', '&lt;')
+    .replaceAll('>', '&gt;')
+    .replaceAll('"', '&quot;')
+    .replaceAll('\'', '&#x27;');
+}
+
```
rofe, line 37: Maybe something for [aem.js](https://github.com/adobe/aem-lib)?

head.html
andreituicu: Good point! Done in [1690e40](https://github.com/adobe/aem-boilerplate/pull/416/commits/1690e4063ccb198de7829d39d4397f77901278d5)

andreituicu:
@rofe I agree with your suggestions of moving different changes into `aem-lib` and `aem-block-collection`.
I created this PR mostly as a consolidated view to collect feedback and discuss whether this is a direction we'd like to go in, or not.

I'm not 100% convinced it is simple enough for developers (which I might be wrong) when they should escape, when they should sanitize and when are both needed.

Using any of them will make the linter happy, but depending on context they aren't both secure, or said differently, each offers protection in specific cases. 

I'm not sure if it would be a better approach to just keeping it simple by saying that any use of unsafe API is unsafe.
The problem then becomes creating big structures with just `document.createElement`, `document.appendChild` and `document.setAttribute` and fragments always need to disable the linter rules, like: https://github.com/adobe/aem-boilerplate/pull/415/files#diff-7fdbc35e8d9d1b49261e0470a1900621a5eb2508a2818a9ee18e2e60514fae4eR26

andreituicu:
Interesting results `1.6` LCP, that's really at the edge. I got `0.9` LCP on pagespeed: https://pagespeed.web.dev/analysis/https-xss-lint-dompurify--helix-project-boilerplate--andreituicu-aem-live/f2hqca9aqv?form_factor=mobile

andreituicu:
Deep PSI said: `929 (1010 ± 139.8)` for LCP.

https://labs.aem.live/tools/deep-psi/deep-psi.html?url1=https%3A%2F%2Fxss-lint-dompurify--helix-project-boilerplate--andreituicu.aem.live%2F&url2=

PR #443, opened 2025-01-10
change collectbase to /.rum/

dkuntze:

This change merely adjusts the collectBaseURL to use the window.origin and thus not the rum.hlx.page domain.

- https://main--aem-boilerplate--adobe.aem.live/
- https://rummy--aem-boilerplate--adobe.aem.live/

### DIFF
--- a/scripts/aem.js
+++ b/scripts/aem.js
@@ -70,7 +70,7 @@ function sampleRUM(checkpoint, data) {
         });
 
         sampleRUM.baseURL = sampleRUM.baseURL || new URL(window.RUM_BASE || '/', new URL('https://rum.hlx.page'));
-        sampleRUM.collectBaseURL = sampleRUM.collectBaseURL || sampleRUM.baseURL;
+        sampleRUM.collectBaseURL = sampleRUM.collectBaseURL || new URL(window.origin);
         sampleRUM.sendPing = (ck, time, pingData = {}) => {
           // eslint-disable-next-line max-len, object-curly-newline
           const rumData = JSON.stringify({


kptdobe (approved):

kptdobe (changes requested):
Ok with the change but aem.js is coming from https://github.com/adobe/aem-lib which brings along the `sampleRUM` function from https://github.com/adobe/helix-rum-js. This is where the change needs to happen!

dkuntze:
> Ok with the change but aem.js is coming from https://github.com/adobe/aem-lib which brings along the `sampleRUM` function from https://github.com/adobe/helix-rum-js. This is where the change needs to happen!

I can change it over there as well, but again, I don't have push perms on that repo.

trieloff (changes requested):
That's wrong, but @kptdobe's suggestion is also wrong, as `aem.js` gets the update pushed from `helix-rum-js`. We need to figure out if we can set `sampleRUM.collectBaseURL` from the outside, and if not add a new variable like `window.RUM_COLLECT` in `helix-rum-js` and set it here in the boilerplate.

dkuntze:
> That's wrong, but @kptdobe's suggestion is also wrong, as `aem.js` gets the update pushed from `helix-rum-js`. We need to figure out if we can set `sampleRUM.collectBaseURL` from the outside, and if not add a new variable like `window.RUM_COLLECT` in `helix-rum-js` and set it here in the boilerplate.

I'm not sure how these are pushed/merged as they are not identical right now (`sampleRUM`). Relatively minor differences, but they are not identical.

trieloff:
`aem-lib` does not allow auto-updates of `helix-rum-js` https://github.com/adobe/aem-lib/blob/main/.renovaterc.json we should change that.

trieloff:
see https://github.com/adobe/aem-lib/pull/124

PR #445, opened 2025-01-10
Quote Stylelint File Globs

### LINKED ISSUE #444: Stylelint Globs need to be quoted
See, https://stylelint.io/user-guide/cli/#usage-examples

Found that if they aren't quoted, files in sub-folders can be skipped. .e.g a file at `blocks/myblock/subfolder/styles.css` is not checked without having the glob quoted.

PR coming

shsteimer:
Fix #444

Test URLs:
- After: https://quote-stylelint-globs--aem-boilerplate--adobe.aem.live/

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "description": "Starter project for Adobe Helix",
   "scripts": {
     "lint:js": "eslint .",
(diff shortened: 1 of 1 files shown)


kptdobe (approved):
Thanks @shsteimer

PR #466, opened 2025-04-05
fix: prefer rum sample rate 1 if param set

maxakuru:
Test URLs:
- Before: https://main--aem-boilerplate--adobe.aem.live/?rum=on
- After: https://rum-param--aem-boilerplate--adobe.aem.live/?rum=on


kptdobe (changes requested):
This line is coming from the https://github.com/adobe/helix-rum-js project (see https://github.com/adobe/helix-rum-js/blob/main/src/index.js#L21-L24). Needs to be fixed there.

PR #470, opened 2025-04-07
Add lint:fix script to package.json

alexcarol:
As autofixing is a common need of all projects i figured it should be implemented in the boilerplate.

Test URLs:
- Before: https://main--aem-boilerplate--adobe.aem.live
- After: https://lint-fix--aem-boilerplate--alexcarol.aem.live


kptdobe (approved):
oh.. I never took the time to try to find the missing `--` trick... Cool, thanks @alexcarol!

PR #493, opened 2025-06-11
feat: Content Security Policy: strict-dynamic + (cached) nonce

andreituicu:
Set the default Content Security Policy for new projects to: https://www.aem.live/docs/csp-strict-dynamic-cached-nonce

Test URLs:
- After: https://csp--aem-boilerplate--andreituicu.aem.live/

CC @davidnuescheler , @rofe -> just because I can't ask for reviews in Github, not having permissions on the boilerplate repo

### DIFF
--- a/404.html
+++ b/404.html
@@ -2,15 +2,20 @@
 <html>
 
 <head>
+  <meta
+    http-equiv="Content-Security-Policy"
+    content="script-src 'nonce-aem' 'strict-dynamic'; base-uri 'self'; object-src 'none';"
+    move-as-header="true"
+  >
   <title>Page not found</title>
-  <script type="text/javascript">
+  <script nonce="aem" type="text/javascript">
     window.isErrorPage = true;
     window.errorCode = '404';
   </script>
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta property="og:title" content="Page not found">
-  <script src="/scripts/scripts.js" type="module"></script>
-  <script type="module">
+  <script nonce="aem" src="/scripts/scripts.js" type="module"></script>
+  <script nonce="aem" type="module">
     window.addEventListener('load', () => {
       if (document.referrer) {
         const { origin, pathname } = new URL(document.referrer);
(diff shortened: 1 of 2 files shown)


davidnuescheler (approved):

404.html
```diff
@@ -2,15 +2,20 @@
 <html>
 
 <head>
+  <meta
+    http-equiv="Content-Security-Policy"
+    content="script-src 'nonce-aem' 'strict-dynamic'; base-uri 'self'; object-src 'none';"
+    move-as-header="true"
```
rofe: move _to_ header? 🤔

rofe (approved):

404.html
andreituicu: The HTML Pipeline is looking at the moment at this exact attribute (`move-as-header`): https://github.com/adobe/helix-html-pipeline/blob/main/src/steps/csp.js#L171-L178

In my mind it sounded ok that it is configured "as a meta tag" and but the HTML pipeline moves it "as a header". Similar to the `import a as b from module` syntax in JS. 🙂 To me, they both sound ok.

I could rename it if you'd like. It would need to be done in a backwards compatible way first and existing customers that use it migrated. At this stage it is, of course, possible since we know exactly which sites use it.
Please let me know what you think!

andreituicu:
@davidnuescheler , @rofe thank you very much for the reviews! I don't have write permissions to this repo, so I would let you merge it when it's ready.

andreituicu:
@davidnuescheler as discussed in the hackathon, I added backwards compatibility directives for browsers that don't support `strict-dynamic`(https://caniuse.com/?search=strict-dynamic). 
*  browsers that support `strict-dynamic` will use:  `script-src 'nonce-aem' 'strict-dynamic'; base-uri 'self'; object-src 'none';`
* browsers that don't support `strict-dynamic` will use: `script-src 'nonce-aem' 'unsafe-inline' http: https:; base-uri 'self'; object-src 'none';`

as far as I can see the main browsers that don't support it are: Safari prior to 15.4.

404.html
andreituicu: `move-as-header` was renamed to `move-to-http-header` in [15caf50](https://github.com/adobe/aem-boilerplate/pull/493/commits/15caf50a2741227203058f9926d1b665198429b2) .

PR #518, opened 2025-08-07
chore: refactor for more compatible media queries

davidnuescheler:
https://no-range-mq--aem-boilerplate--adobe.aem.live/

there is a bit of a conflict with our guidance to not change the default linter configs https://www.aem.live/docs/dev-collab-and-good-practices#linting and the default stylelint config to use media query range syntax. the range syntax is relatively new and has only 92% support https://caniuse.com/css-media-range-syntax

the failure for browsers that don't support it, in our "mobile first" setup really means that only affects desktop (ish) browsers, and they would usually just see the mobile styling.
in cases where js is used via media queries or viewport width dimensions there could be further issues. 

i am not entirely sure what's the right thing to do here, so i am just raising the PR so we have a place to see the impact and debate.

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,3 +1,6 @@
 {
-  "extends": ["stylelint-config-standard"]
+  "extends": ["stylelint-config-standard"],
+  "rules": {
+    "media-feature-range-notation": "prefix"
+  }
 }
--- a/blocks/columns/columns.css
+++ b/blocks/columns/columns.css
@@ -19,7 +19,7 @@
   display: block;
 }
 
-@media (width >= 900px) {
+@media (min-width: 900px) {
   .columns > div {
     align-items: center;
     flex-direction: unset;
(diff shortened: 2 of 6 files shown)


kptdobe:
Most of the non supportive devices seems to be mobile anyway (opera mini and android).

I just got so used to the range syntax... but I agree, this is a dev preference.

fkakatie:
I don't think this PR shouldn't merge.

As far as I can tell, old browsers without range syntax support are predominantly mobile browsers (Opera Mini, old Android). In a mobile-first setup, those devices get base styles, which are mobile styles.

Non-mobile browsers without support is like ~0.5-1% (ancient Safari/Firefox). I don't know if that's any different than the percentage we already decided wasn't worth supporting when dropping IE.

This PR overrides linter defaults to accommodate ~7-8% of users, but isn't most of that cohort NOT actually broken? Just mobile browsers correctly receiving mobile styles?

I don't think it's worth it. Plus, as of this comment, media query range syntax support is up to a whopping 93.96%. Which is almost 94%. Which is ALMOST 95%.

PR #563, opened 2025-12-04
fix: prevent duplicate hero element creation when block is used

sirugh:
### [Related/downstream change](https://github.com/hlxsites/aem-boilerplate-commerce/pull/979)
----
The boilerplate assumes that if an image and h1 are on the page, we should "auto-block" the hero block. But we _also_ have a literal `hero` block, so auto-blocking is not necessary. We should gracefully handle both cases.

This change introduces a fix which ensures that if the block instrumentation exists in the page source, it will NOT auto-block.

In this PR, both manners of implementing a hero should work - using the block or just using an image + h1.

With block
image
Without block
image

Test URLs:
- Before with hero block: https://main--aem-boilerplate-commerce--hlxsites.aem.live/drafts/rugh/double-hero-debug
- Fix with hero block: https://dupe-hero-fix--aem-boilerplate-commerce--hlxsites.aem.live/drafts/rugh/double-hero-debug
- Before with just image/h1: https://main--aem-boilerplate-commerce--hlxsites.aem.live/drafts/rugh/double-hero-debug-2
- Fix with just image/h1: https://dupe-hero-fix--aem-boilerplate-commerce--hlxsites.aem.live/drafts/rugh/double-hero-debug-2

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -22,6 +22,10 @@ function buildHeroBlock(main) {
   const picture = main.querySelector('picture');
   // eslint-disable-next-line no-bitwise
   if (h1 && picture && (h1.compareDocumentPosition(picture) & Node.DOCUMENT_POSITION_PRECEDING)) {
+    // Check if h1 or picture is already inside a hero block
+    if (h1.closest('.hero') || picture.closest('.hero')) {
+      return; // Don't create a duplicate hero block
+    }
     const section = document.createElement('div');
     section.append(buildBlock('hero', { elems: [picture, h1] }));
     main.prepend(section);


scripts/scripts.js
```diff
@@ -22,6 +22,10 @@ function buildHeroBlock(main) {
   const picture = main.querySelector('picture');
   // eslint-disable-next-line no-bitwise
   if (h1 && picture && (h1.compareDocumentPosition(picture) & Node.DOCUMENT_POSITION_PRECEDING)) {
+    // Check if h1 or picture is already inside a hero block
+    if (h1.closest('.hero') || picture.closest('.hero')) {
```
bstopp, line 26: Should this check for `.block` instead of `.hero` - just in case implementations have other non-hero blocks that contain H1 and a pciture?

```suggestion
    if (h1.closest('.block') || picture.closest('.block')) {
```

sirugh: People can certainly do what they want, but in this case, it seems like the implementation either expects the first image and h1 to be the hero image/text, OR for a hero block to exist on the page.

Also, I don't see `.block` used in my source so not sure where that comes from 

Image

bstopp: ah yes, that's a good point, it shows up after the `decroate....` runs

trieloff (commented):
looks reasonable.

kptdobe (approved):
Makes sense. Thanks @sirugh

PR #582, opened 2026-02-03
fix(lib): update scripts/aem.js to aem.js@2.10.6

trieloff:
Test URL: https://update-lib-aem-patch-2-10-6--aem-boilerplate--adobe.aem.live/
Release Notes: https://github.com/adobe/aem-lib/releases/tag/v2.10.6


fkakatie (approved):
**Behavior impact:**

- Fixes previously broken cases where absolute `src` images got rewritten to the current site’s path.
- Should be neutral for normal relative images.
- Edge cases like `//cdn…`, `data:`, or `blob:` URLs still aren’t handled specially.

Issue #119, opened 2022-08-09
Add ability to specify link target


sdmcraft:
### Expected Behaviour
Page authors should have the ability to specify the [link target](https://www.w3schools.com/tags/att_a_target.asp) i.e. whether it opens in the same tab/ new tab etc in the authoring document.
 
### Actual Behaviour
Currently there's no such ability.  Developers need to implement custom logic for achieving this. For e.g. https://github.com/hlxsites/mmm-af/commit/9dba40001d9a9c461d92c4a1e13b6c2cc7b5e1b9

### Reproduce Scenario (including but not limited to)
Not Applicable

#### Steps to Reproduce
Not Applicable

#### Platform and Version
All

#### Sample Code that illustrates the problem
Not Applicable

#### Logs taken while reproducing problem
Not Applicable

trieloff:
I don't think this is nearly as common enough to be left to authors. In most cases, sites can formulate rules such as "all external links should open in a new window except for these hostnames", so that no one-off decisions by authors are needed.

sdmcraft:
@trieloff , I logged this ticket based on a [discussion here](https://cq-dev.slack.com/archives/C02DL4SSXCN/p1659166222212469) wherein @davidnuescheler mentioned that this has come more than once and could be considered for inclusion in the bolierplate. If we feel that isn's the case or we need to reconsider, I can close this. Else perhaps an [implementation like this](https://github.com/hlxsites/mmm-af/commit/9dba40001d9a9c461d92c4a1e13b6c2cc7b5e1b9) can be ported here,

ramboz:
So the question seems to be if we can find valid use-cases for internal links to open in new tabs consistently.

From the top of my head, I can think of 2 use cases that could benefit from this:
1. Links to pdf, doc, xls, etc. would benefit from being opened in a new tab all the time as they represent a complete change of context (from an accessibility point-of-view, and would need to be combined with https://www.w3.org/TR/WCAG20-TECHS/G201.html)
2. Links to contextual documentation or glossaries could be opened in a named window/modal with `#` anchors to scroll to the right section immediately. I think that wouldn't work great with the proposed `#_blank` option, as we'd need to combine 2 anchors

And then there is a 3rd… more gray area one:

3. Links to code samples, which may include css/js and would thus be hard to differentiate from the ones you'd want to download forcibly. That's probably more project-specific so harder to define a default behavior. But 2. could also address this.

fkakatie:
[...]

PR #584, opened 2026-02-04
feat: open file links in new tabs automatically

fkakatie:
Automatically opens links to file types (PDF, DOC, media, etc.) in new tabs.

Runs in `loadDelayed` as a progressive enhancement (good example of the kind of thing that belongs there). Projects can easily extend or override with their own link decoration rules. No breaking changes.

Addresses #119

Test URLs:
- After: https://file-target-blank--aem-boilerplate--adobe.aem.live/

### DIFF
--- a/scripts/delayed.js
+++ b/scripts/delayed.js
@@ -9,12 +9,13 @@ function decorateFileLinks(doc) {
     '.doc', '.docx', '.xls', '.xlsx', '.ppt', '.pptx',
     '.pdf', '.zip', '.mp4', '.mov', '.avi',
   ];
-  
+
   doc.querySelectorAll('a[href]').forEach((link) => {
     try {
       const url = new URL(link.href);
-      const hasFileExtension = fileExtensions.some((ext) => url.pathname.toLowerCase().endsWith(ext));
-      
+      const pathname = url.pathname.toLowerCase();
+      const hasFileExtension = fileExtensions.some((ext) => pathname.endsWith(ext));
+
       if (hasFileExtension && !link.hasAttribute('target')) {
         link.setAttribute('target', '_blank');
       }
@@ -24,4 +25,4 @@ function decorateFileLinks(doc) {
   });
 }
 
-decorateFileLinks(doc);
+decorateFileLinks(document);
--- a/scripts/delayed.js
+++ b/scripts/delayed.js
@@ -1,1 +1,27 @@
 // add delayed functionality here
+
+/**
+ * Decorates links to file types to open in new tabs.
+ * @param {Element} doc The container element
+ */
+function decorateFileLinks(doc) {
+  const fileExtensions = [
+    '.doc', '.docx', '.xls', '.xlsx', '.ppt', '.pptx',
+    '.pdf', '.zip', '.mp4', '.mov', '.avi',
+  ];
+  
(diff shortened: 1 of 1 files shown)


fkakatie:
Note: I don't think this can be fixed in boilerplate without violating [WCAG 2.0 G201](https://www.w3.org/TR/WCAG20-TECHS/G201.html) (advance warning when opening new windows) OR our "no hard-coded UI text" guideline.

**Options:**
1. PDF only: Narrower scope, more expected behavior, still violates G201
2. Ship anyway: Document the trade-off, let maintainers decide
3. Close as out of scope: Provide example code, let projects implement with proper localized text

trieloff:
I'm firmly in "ship anyway" camp.

trieloff (approved):
Approved after linting passes.

rofe:
Why is this needed? Wouldn't this leave you stranded with an empty tab, at least on mobile?

ramboz:
I think modern browsers already handle downloads reasonably well, and users also have established browser- and OS-level affordances to deterministically choose how links behave (open, download, save as, etc.).

Forcing download behavior via markup or JS can actually conflict with WCAG 2.1/2.2, particularly around:

- Predictability (3.2.1 / 3.2.2)
- User control and expectations

So I’m not a big fan of enforcing this. It feels like an opinionated approach that is likely to cause more issues down the road than it will solve.

From an accessibility and UX standpoint, I’d argue that clear disclosure is the best practice, not forced behavior. For external links, there are already patterns like a suffix icon + accessible text; I’d apply the same approach here for downloadable assets.

I'd suggest the following:

1. Suffix the link with a clear visual indicator. Either a simple download icon, a downward arrow, or (ideally) a file-type indicator (PDF, DOCX, etc.).
2. Provide screen-reader-only text describing the file type, e.g. “(PDF)” or “(DOCX)”, included in the accessible name of the suffix.
3. Optionally include file size when available. This is a WCAG-aligned best practice, as it helps users anticipate download time and data usage.
4. If the link opens a new tab/window, disclose that explicitly. At least in screen-reader-only text.

This keeps behavior predictable, respects user preferences, and ensures assistive technology users get the same level of information as sighted users—without relying on tooltips or browser-specific quirks.

fkakatie:
Thanks @ramboz for the thorough accessibility analysis and to everyone who indulged me in this discussion. The consensus that disclosure > forced behavior aligns well with boilerplate's philosophy of providing flexible infrastructure rather than opinionated solutions. Closing this PR and #119 as **out of scope for boilerplate**.

## Why This Doesn't Belong Here

While external/file link decoration is a valid project need, it's too opinionated for boilerplate because:

1. **Not universal**: Different projects have different rules for when/how to open links
2. **Localization required**: WCAG-compliant warnings need project-specific text
3. **Design-dependent**: Visual indicators vary by brand/design system
4. **Behavior opinions differ**: As @ramboz noted, forcing behavior conflicts with user preferences and WCAG predictability principles

This feels more like project-specific functionality that doesn't benefit all downstream projects universally.

## Alternative: Block Party 🎉 
If anything, this (or something like this) might be a good candidate for a [Block Party](https://www.aem.live/developer/block-party/) contribution, where projects can discover and adopt it if needed.

PR #586, opened 2026-02-05
feat: use requestIdleCallback for delayed script loading

fkakatie:
Addresses discussion in #583 and relates to #376 and #408

This PR implements the `requestIdleCallback` approach proposed by @ramboz in discussion #583.

## Changes

- Modified `loadDelayed()` to use `requestIdleCallback` when available
- Loads delayed scripts only when CPU has spare capacity or falls back to existing 3-second `setTimeout` for browsers without support ([only ~82% coverage](https://caniuse.com/?search=requestIdleCallback) as of this comment)
- Prevents INP and TBT issues

## Benefits

- **Better performance**: No artificial delays or user interaction requirements
- **Predictable behavior**: Loads for all visitors
- **Progressive enhancement**: Modern browsers get optimal loading, older browsers get current behavior
- **Non-breaking**: Consent handling remains separate, scripts in delayed.js unaffected

Test URLs:
- After: https://idle-delayed-loading--aem-boilerplate--adobe.aem.live/

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -139,7 +139,14 @@ async function loadLazy(doc) {
  */
 function loadDelayed() {
   // eslint-disable-next-line import/no-cycle
-  window.setTimeout(() => import('./delayed.js'), 3000);
+  const importDelayed = () => import('./delayed.js');
+
+  if ('requestIdleCallback' in window) {
+    // prevents INP/TBT issues by only loading when CPU has capacity
+    window.requestIdleCallback(importDelayed);
+  } else {
+    window.setTimeout(importDelayed, 3000); // fallback 3-second timeout
+  }
   // load anything that can be postponed to the latest here
 }
 
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -143,7 +143,7 @@ function loadDelayed() {
 
   if ('requestIdleCallback' in window) {
     // prevents INP/TBT issues by only loading when CPU has capacity
-    window.requestIdleCallback(importDelayed);
+    window.requestIdleCallback(importDelayed, { timeout: 3000 });
   } else {
     window.setTimeout(importDelayed, 3000); // fallback 3-second timeout
   }


scripts/scripts.js
```diff
@@ -139,7 +139,14 @@ async function loadLazy(doc) {
  */
 function loadDelayed() {
   // eslint-disable-next-line import/no-cycle
-  window.setTimeout(() => import('./delayed.js'), 3000);
+  const importDelayed = () => import('./delayed.js');
+
+  if ('requestIdleCallback' in window) {
+    // prevents INP/TBT issues by only loading when CPU has capacity
+    window.requestIdleCallback(importDelayed);
```
ramboz: I'd also use a timeout on the idle callback just to be safe.
```suggestion
    window.requestIdleCallback(importDelayed, { timeout: 3000 });
```

ramboz (approved):

shsteimer:
I love this idea, but think we may want to consider ensuring all the promises in lazy (header, footer, fonts, lazy-styles.css) resolve before delayed starts if we do this. Currently, with the 3 second delay, those lacking awaits aren't an issue. But I believe it's possible for the cpu to be idle while those promises are still pending, in which case `delayed.js` would start loading before the footer or fonts are fully loaded.

I don't feel super strongly about it, maybe it's still a non-issue, but wanted to at least raise it for consideration.

ramboz (approved):

ramboz:
@shsteimer I don't think the new implementation really changes anything compared to the old one, as it relates to the promise order and awaiting.

I'd argue that this could indeed be an edge case if you do enhance existing elements in the delayed phase without properly awaiting their initial decoration in the lazy phase. There could indeed be race conditions on some devices. But I also think this can easily be solved by properly waiting the promises in the `loadLazy` at the project level based on the exact needs of the project and it's not something that would be needed for 80% of the projects, so I wouldn't really address it in the boilerplate.

shsteimer:
Using this with a customer now, what I've found is that since the stuff they load in `delayed.js` causes some TBT (pretty normal IME) and SI (consent popup message). These drop the lighthouse scores from roughly 100->85, in this particular case. Neither of these are CWV, so CWV should still be all green once we go live. And in some sense you could say this is "fake degradation" since the TBT and SI are there even with the 3 second delay, they just occur outside the window that google uses to measure LHS.

it's one data point, and not sure it should change our approach here, but I think at least worth considering.

ramboz:
@shsteimer Fully agree with you here. Moving around the failing metrics outside the measuring window ultimately also doesn't help the end user. Thes still face the same problem, just at a different time in the page lifecycle.

- TBT: if you have access to the code that triggers the TBT, breaking up the long task should help. `setTimeout`/`requestIdleCallback`/`scheduler.yield()` could likely all be good options there.
- SI: this will likely be a moot point to optimize. As long as a good chunk of pixels move on the screen, it will flag it as SI. SI by itself is not necessarily bad. CLS is the real concern usually.

PR #639, opened 2026-06-22
chore(deps): upgrade ESLint to v10 with flat config

tripodsan:
## Summary

- Upgrades `eslint` from `8.57.1` to `^10.5.0`
- Replaces `.eslintrc.js` + `.eslintignore` with `eslint.config.mjs` (flat config, required by ESLint v9+)
- Drops `@babel/eslint-parser` — replaced by `sourceType: 'module'` in `languageOptions`
- Drops `eslint-config-airbnb-base` and the `@eslint/eslintrc` FlatCompat shim — replaced by `@eslint/js` recommended, which is natively compatible with ESLint 10
- Keeps `eslint-plugin-import` for the `import/extensions` rule; all other project-specific rules (`linebreak-style`, `no-param-reassign`) are set directly in `eslint.config.mjs`
- Removes the ESLint v8 pin from `.renovaterc.json`
- Adds `scripts/aem.js` to `ignores` (uses `/* eslint-env */` directive which ESLint 10 no longer supports; file must not be modified per project rules)
- Removes stale `eslint-disable` comments across several files that referenced rules no longer in scope

## Test plan

- [x] `npm run lint` passes with zero errors and zero warnings

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tripodsan:
problem is that https://github.com/airbnb/javascript/issues/2961 still doesn't support eslint9+

rofe (approved):

fkakatie (approved):

PR #641, opened 2026-06-23
feat: Harden CSP protection with Trusted Types

andreituicu:
Fix [SITES-46900](https://jira.corp.adobe.com/browse/SITES-46900) -> We can use the JIRA for internal conversations

Test URLs:
- After: https://csp-update--aem-boilerplate--adobe.aem.live/

# Changes

1. Add `frame-src` directive to the CSP -> make sure that the src attribute of iframes can only be a `https:` url or self, so `data:` protocols and similar are deactivated.
2. Enable trusted types, the new baseline feature for browsers: https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API 
3. Create a default Trusted Types policy which helps cover remaining DOM based edge cases that cannot be covered with the cached nonce approach.

### DIFF
--- a/head.html
+++ b/head.html
@@ -1,8 +1,8 @@
 <meta
   http-equiv="Content-Security-Policy"
-  content="script-src 'nonce-aem' 'strict-dynamic' 'unsafe-inline' http: https:; base-uri 'self'; object-src 'none';"
+  content="script-src 'nonce-aem' 'strict-dynamic' 'unsafe-inline' http: https:; base-uri 'self'; object-src 'none'; frame-src 'self' https:; require-trusted-types-for 'script';"
   move-to-http-header="true"
->
+/>
 <meta name="viewport" content="width=device-width, initial-scale=1"/>
 <script nonce="aem" src="/scripts/aem.js" type="module"></script>
 <script nonce="aem" src="/scripts/scripts.js" type="module"></script>
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -12,6 +12,64 @@ import {
   buildBlock,
 } from './aem.js';
 
+if (window.trustedTypes && window.trustedTypes.createPolicy) {
+  const innerTT = window.trustedTypes.createPolicy('tt-inner', {
+    // avoid stack overflow
+    createHTML: (s) => s,
+  });
+
+  window.trustedTypes.createPolicy('default', {
+    /**
+     * All HTML creation goes through this function, so we can sanitize it for known attack vectors.
+     * @param {string} input The HTML input string
+     * @param {string} type The type of HTML being created
+     * @param {string} sink The sink where the HTML will be used
+     * @returns {undefined|string} The sanitized HTML string or undefined if the input is unsafe
+     */
+    createHTML: (input, type, sink) => {
+      // DOMPurify or similar sanitization library may be implemented here
+      // if a harder policy is desired, with a tradeoff on performance.
+      let processedInput = input;
+      if (/srcdoc\s*=/i.test(processedInput)) {
+        const doc = new DOMParser().parseFromString(innerTT.createHTML(processedInput), 'text/html');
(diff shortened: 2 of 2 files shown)


head.html
```diff
@@ -1,8 +1,8 @@
 <meta
   http-equiv="Content-Security-Policy"
-  content="script-src 'nonce-aem' 'strict-dynamic' 'unsafe-inline' http: https:; base-uri 'self'; object-src 'none';"
+  content="script-src 'nonce-aem' 'strict-dynamic' 'unsafe-inline' http: https:; base-uri 'self'; object-src 'none'; frame-src 'self' https:; require-trusted-types-for 'script';"
```
lkrapf, line 3: 'unsafe-inline', 'http:', and 'https:' only matter for browsers that don't support 'strict-dynamic', but those browsers would likely break on AEM's dynamic script loading anyway — so they seem redundant here.

lkrapf:
LGTM, neither claude nor I (FWIW ;) ) could break it.

lkrapf (approved):

head.html
andreituicu: Thank you very much!

There are two reasons for still keeping the `'unsafe-inline' http: https:` portion:
1. We did have a customer reaching out last year that they had users reporting broken website on an old safari version, which supported dynamic script loading, but not `strict-dynamic`
2. Google's Lighthouse Report will still show a warning if these directives are not present for backwards compatibility purposes

So because of these I thought the best course of action would be to wait until Google removes that recommendation from their report.

Screenshot 2026-06-24 at 09 32 39

davidnuescheler (approved):

scripts/scripts.js
```diff
@@ -12,6 +12,31 @@ import {
   buildBlock,
 } from './aem.js';
 
+if (window.trustedTypes && window.trustedTypes.createPolicy) {
```
rofe, line 15:
```suggestion
if (window.trustedTypes?.createPolicy) {
```

scripts/scripts.js
```diff
@@ -12,6 +12,31 @@ import {
...
+  const innerTT = window.trustedTypes.createPolicy('tt-inner', {
+    createHTML: (s) => s, // avoid stack overflow
+  });
+
+  window.trustedTypes.createPolicy('default', {
+    createHTML: (input, type, sink) => {
+      let processedInput = input;
+      if (/srcdoc\s*=/i.test(processedInput)) {
+        const doc = new DOMParser().parseFromString(innerTT.createHTML(processedInput), 'text/html');
+        doc.querySelectorAll('[srcdoc]').forEach((el) => el.removeAttribute('srcdoc'));
```
rofe: Since `srcdoc` s an `iframe`-only attribute, we should probably restrict the selector to `iframe` elements for performance reasons:
```suggestion
        doc.querySelectorAll('iframe[srcdoc]').forEach((el) => el.removeAttribute('srcdoc'));
```

rofe (changes requested):

scripts/scripts.js
andreituicu: On this particular suggestion, @davidnuescheler had mentioned some time ago that he would not like to introduce the optional operator in the boilerplate code.
Given that there is no other example, I'm guessing this restriction is still in place.
From my point of view I'm ok to both use it, or continue the restriction; I have no preference.

rofe: Well, that must have been in 2020 when we first introduced optional chaining in Express and it broke for some users with old Safaris... that was 6 years ago. There is [documented support for optioanl chaining](https://caniuse.com/mdn-javascript_operators_optional_chaining) in all major browsers since 2020, and I have been using it many times for its elegance and brevity.

But I won't block this PR because this line looks old-fashioned to me :) 

cc @davidnuescheler

rofe (approved):

PR #649, opened 2026-07-22
fix(widget): apply shell config before widget hydration

davidnuescheler:
## Summary

- Reorders widget block decoration so metadata, block classes, and wrapper/container shell classes are applied before fetching widget HTML and loading widget JS
- Extracts shell setup into `applyWidgetShell()` so widgets can read `dataset`, name classes, and parent selectors during their decorate function

## Context

Previously the block followed a "hydrate first, attach metadata and shell classes after" pattern. That works for widgets that don't need config at decorate time, but breaks widgets that read query-param `data-*` attributes, widget name classes, or `{name}-wrapper` / `{name}-container` parents when they initialize.

## Test plan

- [ ] Add or use a widget whose `decorate()` reads `el.dataset` or parent shell classes
- [ ] Author a page with a widget link including query params (e.g. `/widgets/demo/demo.html?theme=dark`)
- [ ] Verify config is available when the widget decorate function runs
- [ ] Verify resulting DOM still has correct `{name}`, `{name}-wrapper`, and `{name}-container` classes after load
- [ ] Preview: https://fix-widget-shell-before-hydrate--aem-boilerplate--adobe.aem.page/

Made with [Cursor](https://cursor.com)

### DIFF
--- a/blocks/widget/widget.js
+++ b/blocks/widget/widget.js
@@ -24,6 +24,34 @@ function widgetUrl(widgetPath, widgetName, extension) {
 }
 
 /**
+ * Applies widget metadata, block classes, and section shell classes.
+ * Must run before widget HTML/JS load so decorate can read config from the DOM.
+ * @param {Element} widget The widget block element
+ * @param {HTMLAnchorElement} source The authored widget link
+ * @param {string} widgetName Widget file name without extension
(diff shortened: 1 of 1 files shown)


fkakatie (approved):
this actually makes much more sense to my brain.

PR #658, opened 2026-08-13
feat: sync to block collection

fkakatie:
## Why

[`aem-block-collection`](https://github.com/adobe/aem-block-collection) doesn't stay in sync with `aem-boilerplate` the way `aem-boilerplate` stays in sync with [`aem-lib`](https://github.com/adobe/aem-lib), requiring manually pulling in the latest changes.

## What this does

Adds a GitHub Action to `aem-boilerplate` that, on every push to `main`, checks whether `scripts/aem.js`, `scripts/scripts.js`, or `head.html` changed. 
- If so, it opens a PR against `aem-block-collection` with those files updated (the same pattern `aem-lib` already uses to keep `aem-boilerplate`'s `scripts/aem.js` current, just one repo further downstream). 
- If none of those files changed, nothing happens.

The scope is intentionally narrow: just the three files that have actually caused drift and regressions so far. It doesn't touch blocks or docs and it's guarded so it can never accidentally run in a project someone creates from this template (only in `adobe/aem-boilerplate` itself).

## Before this goes live

- A repo secret needs to be added to `adobe/aem-boilerplate`. 
- A manual test run (`workflow_dispatch`) against a real past commit before letting it run unattended on every push, just to confirm the PR it opens looks right.
- Confirm scope. Right now it's just these three files, not the shared blocks (`cards`, `header`, `hero`, etc.), since block-collection may have intentionally diverged some of those.

## Test URLs:
- After: https://sync-block-collection--aem-boilerplate--adobe.aem.live/

### DIFF
--- a/.github/workflows/cleanup-on-create.yaml
+++ b/.github/workflows/cleanup-on-create.yaml
@@ -23,9 +23,11 @@ jobs:
         with:
           node-version: 24
       - name: Remove Helper Files
-        run: | 
+        run: |
           rm -rf \
             .github/workflows/cleanup-on-create.yaml \
+            .github/workflows/sync-block-collection.yaml \
+            sync-block-collection.sh \
             .renovaterc.json \
             CHANGELOG.md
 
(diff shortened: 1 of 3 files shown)


rofe (approved):
❤️

PR #665, opened 2026-09-18
fix(lib): update scripts/aem.js to aem.js@3.1.9

trieloff:
Test URL: https://update-lib-aem-patch-3-1-9--aem-boilerplate--adobe.aem.live/
Release Notes: https://github.com/adobe/aem-lib/releases/tag/v3.1.9


ramboz (approved):
## Summary
This pulls in aem-lib 3.1.9 and adds a `dataFromEventObj` helper so unhandled promise rejections whose reason is a DOM `Event` (the `el.onerror = reject` pattern from `loadScript`/`loadCSS`) get a real locator in the RUM beacon instead of a useless serialized Event. Small, single-file, matches the repo's established pattern for these lib bumps (same title format as the last several: 2.7.x, 2.8.0, now 3.1.9).

## Strengths
- The fallback chain in `dataFromEventObj` is genuinely careful: tag name first, then `src`/`href` filtered to strings only (with a comment explaining why, since `href` on an SVG element is an `SVGAnimatedString`, not a string), then falls back to `event.type`, then caps at 200 chars so a `data:` URL can't bloat the beacon. That's a real edge case caught, not just happy-path code.
- It's a strict improvement over the previous behavior: an Event reason used to fall through to `target: reason`, which serializes to nothing useful on the wire. Now it's an actual locator.
- CI is green (build, aem-psi-check, Kodiak, CLA) and the PR links the upstream release notes and a live test URL, consistent with how this repo has always handled these bumps.

## Verdict
**Ready to merge:** Yes
**Reasoning:** Defensive, well-scoped vendored update with no behavior change to existing paths and a clear improvement to a previously-useless error case.

---
*Note: reviewed from the diff and PR history alone (no test suite exists in this repo to check against, and this is a routine vendored-lib sync trieloff has done many times before). Nothing else to verify.*

*Automated review.*

